PR #3, opened 2019-01-10
Use numpy

samuelhklumpers:
Replaced previous calculations with their numpy equivalents and rewrote filepath logic to use join rather than string concat.


Smitty1298:
Great work!

PR #18, opened 2019-01-25
Caching

ChrisMiuchiz:


osu-ac/cacher.py
```diff
@@ -11,7 +12,7 @@ class Cacher:
     methods provide cleaner access than passing around a Cacher class.
     """
 
-    conn = sqlite3.connect(PATH_DB)
+    conn = sqlite3.connect(str(PATH_DB))
```
tybug: sqlite3.connet takes in a path like object (see https://docs.python.org/3.7/library/sqlite3.html#sqlite3.connect), which PATH_DB is. No need to cast to str

osu-ac/cacher.py
```diff
@@ -34,7 +35,8 @@ def cache(map_id, user_id, lzma_string, replay_id):
         """
 
         compressed_string = Cacher.compress(lzma_string)
-        Cacher.write("INSERT INTO replays VALUES(?, ?, ?, ?)", [map_id, user_id, compressed_string, replay_id])
+        packed_osr = wtc.compress(compressed_string)
```
tybug: return wtc.compress(lzma_string) from Cacher.compress, I left that empty specifically for this implementation

osu-ac/cacher.py
tybug: I take this back, python3.6 sqlite3 does not support path0-like objects (https://docs.python.org/3.6/library/sqlite3.html#sqlite3.connect), which we definitely want to support.

PR #22, opened 2019-01-30
Create requirements.txt, add new usage to README

Poliwrath:
idk what to put here

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-numpy==1.16.0
-matplotlib==3.0.2
-requests==2.21.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -0,0 +1,5 @@
+numpy==1.16.0
```
tybug: we want to support future versions too so use >= unless there's a good reason not to

Poliwrath: 🎉

tybug:
Thanks! our dev discord is here, if you want to join https://discord.gg/JRW9GH

PR #25, opened 2019-01-31
Update README.md

spook0:
lol

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ As of the v1.0 release, we only attempt to detect the first item in that list - 
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


spook0:
hello

tybug:
For the record, this user was blocked from the org and banned from the discord for spam and pr's that go directly against discussion.

qewc:
Not the biggest fan of spook, but he *does* have a point.

tybug:
He does. Unsure if you are in the discord, but we already discussed providing support for earlier versions by removing the f format strings from the osrparse and compression repos.

qewc:
Yeah im in the discord, f strings look cleaner. However, "{}".format(variable) will do the same job.

PR #26, opened 2019-02-01
image fix, video link update

tybug:


tybug:
force merge because these are simple updates

PR #47, opened 2019-02-03
Add Progress Tracker every 10% 

VINXIS:
Example: https://imbouttaheadout.s-ul.eu/yHf19qWE

Also a print showing the total combinations as well

Sry for Pr clog

### DIFF
--- a/osu-ac/comparer.py
+++ b/osu-ac/comparer.py
@@ -1,6 +1,7 @@
 import itertools
 
 import numpy as np
+import math
 
 from draw import Draw
 from replay import Replay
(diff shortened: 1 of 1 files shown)


osu-ac/comparer.py
```diff
@@ -64,12 +65,16 @@ def compare(self, mode):
...
         elif (mode == "single"):
             iterator = itertools.combinations(self.replays1, 2)
+            total = sum(1 for i in itertools.combinations(self.replays1, 2))
         else:
             raise InvalidArgumentsException("`mode` must be one of 'double' or 'single'")
 
-        print("Starting to compare replays")
+        tenth = math.floor(total / 10)
+        done = 0
+        print("Starting a total of {:d} combinations.".format(total))
```
tybug: can you remove the period here and the other print? none of our other prints have it

osu-ac/comparer.py
```diff
@@ -89,13 +94,21 @@ def compare(self, mode):
 
             for key in results:
                 self._print_result(results[key], key[0], key[1])
+            done += 1
+            rem = done % tenth
```
tybug: make this and the other one a one liner, if(done % tenth == 0)

tybug: oh you can't do that lol because you increment done

samuelhklumpers: you could use enumerate and not have to increment

tybug: yeah wrap an enumerate call around the for loops and use that instead, also your stddevs loop logic is wrong - you never increment inside a loop. Wrap enumerate around the proper loop there; since we print results after all comparisons are done the progress bar will always finish printing there before any results print.

VINXIS:
OK

tybug:
mergeable, not the greatest (haven't quite though it through yet but a separate `progress` class might go a long way to simplifying these multi line print+flush statements). We'll save that for another day, another rewrite, though. Thanks for the pr.

tybug:
merged in febe2737995647668b9807b5c4dbb2d758ea1586, can't for the life of me figure out why it didn't auto close this pr, maybe because I squashed merged and didn't normal merge I don't know

PR #48, opened 2019-02-03
Ability to save visualization as Video

InvisibleSymbol:
Also small changes in draw()
First few frames maybe broken, need to look into it
~shows progress of saving every 10%~
video is saved as "player_name0 vs player_name1.mp4"
Example : 
https://streamable.com/fi3ur

### DIFF
--- a/circleguard/draw.py
+++ b/circleguard/draw.py
@@ -87,14 +87,14 @@ class Draw():
             plot1.set_data(data1[0][i - 100:i], data1[1][i - 100:i])
             plot2.set_data(data2[0][i - 100:i], data2[1][i - 100:i])
             if i != 0 and i % int((total/10)) == 0:
-                print(f"Saved {round(100 * float(i)/float(total))}% already!")
+                print(f"Saved {round(100 * float(i)/float(total))}% of the video")
             return plot1, plot2
 
         setup()
         
         plt.show(block=True)
         plt.close('all')  # Maybe unnecessary
-        if input("Do you want to save the visualization? WARNING! This may take long ") == "y":
+if input("Do you want to save the video? WARNING, this may take a while ")[0] == "y":
             print("started saving video")
             setup()
 
--- a/circleguard/draw.py
+++ b/circleguard/draw.py
@@ -85,8 +85,6 @@ class Draw():
         def update(i):
             plot1.set_data(data1[0][i - 100:i], data1[1][i - 100:i])
             plot2.set_data(data2[0][i - 100:i], data2[1][i - 100:i])
-            if i != 0 and i % int((total/10)) == 0:
-                print(f"Saved {round(100 * float(i)/float(total))}% of the video")
             return plot1, plot2
 
         setup()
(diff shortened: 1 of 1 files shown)


InvisibleSymbol:
oh, ffmpeg is needed in the dir, don't know how to handle that

circleguard/draw.py
```diff
@@ -62,13 +62,21 @@ def run(self):
...
         data2 = np.transpose(data2)
 
-        # create plot for each replay and add legend with player names
-        fig, ax = plt.subplots()
+        total = len(data1[0])  # stationary cause otherwise it would be a pain
 
-        plot1 = plt.plot('x', 'y', "red", animated=True, label=self.replay1.player_name)[0]
-        plot2 = plt.plot('', '', "blue", animated=True, label=self.replay2.player_name)[0]
+        def setup():
+            global plot1, plot2, fig, ax, animation
```
tybug: using global here doesn't sit right with me, leaving this to sam's discretion because he actually knows matplotlib

circleguard/draw.py
```diff
@@ -78,10 +86,26 @@ def init():
         def update(i):
             plot1.set_data(data1[0][i - 100:i], data1[1][i - 100:i])
             plot2.set_data(data2[0][i - 100:i], data2[1][i - 100:i])
+            if i != 0 and i % int((total/10)) == 0:
+                print(f"Saved {round(100 * float(i)/float(total))}% already!")
```
tybug: replace with 
`print(f"Saved {round(100 * float(i)/float(total))}% of the video")`

circleguard/draw.py
```diff
@@ -78,10 +86,26 @@ def init():
...
             return plot1, plot2
 
-        animation = FuncAnimation(fig, update, frames=itr.count(100), init_func=init, blit=True, interval=dt)
+        setup()
+        
         plt.show(block=True)
+        plt.close('all')  # Maybe unnecessary
+        if input("Do you want to save the visualization? WARNING! This may take long ") == "y":
```
tybug: `if input("Do you want to save the video? WARNING, this may take a while ")[0] == "y":`

circleguard/draw.py
InvisibleSymbol: done

circleguard/draw.py
samuelhklumpers: yeah try removing this because we already return animation and store it at the caller, which should be enough

InvisibleSymbol:
indents seems broken, yay

circleguard/draw.py
InvisibleSymbol: global is necessary so ax & co can be accessed outside the function, without global you get an `NameError: name 'ax' is not defined` Error

circleguard/draw.py
InvisibleSymbol: Inserting [0] indeed works, but this will fail if there is no input. I instead choose to use:
`if (answer and answer[0].lower().strip() == "y")`
This is also used in "comparer.py"

tybug:
Closing this issue because we can't delete the developement branch and create the development branch (note the lack of a typo in the latter!) with a dependent PR. Requiring a dependency that we can't easily add through pip or other automatic download means is still an outstanding issue with this pr, though excellent in concept. Feel free to reopen this as either an issue (outlining wanting the ability to save as a video) or a new pr to development if you want to deal with merge conflicts.

PR #49, opened 2019-02-04
 Show amount of received replays with ETA

InvisibleSymbol:
probably not nicely cleaned up
sorry for the weird commits, but should work
Example:


tybug:
Not happy with the way you pass variables through all the methods, which makes for example `enforce_ratelimit` impossible to use with anything but map data. A better solution would be to make Loader an instance initialized once per Anticheat instance, much the same way a Cacher is, and store loaded/total as instance variables there. (Make sure to keep ratelimits counts as static variables though, those are shared across threads)

InvisibleSymbol:
~hopefully correctly~ rewritten

InvisibleSymbol:
It seems that the timeout is shared... Is this unwanted?

tybug:
nice rewrite. Not entirely sure what you mean by timeout being shared, but each instance should be getting ratelimited, yes. Because we don't receive a "time left until ratelimit refresh" from the api, we have to estimate per thread the time left.

InvisibleSymbol:
Sorry, somehow I thought that on the master branch each thread had his own timeout...
Basically what now happens is that each thread waits the same amount of time, causing a some sort of request spam when the timeout finishes. But seems like this was always the case...

tybug:
This is okay as it stands, but it would be better to move the logic of creating OnlineReplays to the Loader instead of calling OnlineReplay.from_user_info. That way we wouldn't have to pass the Loader around. I'll do that part later today if you don't want to have first stab at it.

InvisibleSymbol:
yea, it would be nice if you could try it before me. Thanks for helping me.

tybug:
Merged in #53 instead for convoluted reasons that include not being able to specify a co-author in squash commits

PR #50, opened 2019-02-04
Exceptions

tybug:
super clean api exception handling and exception restructure. Should catch all request exceptions as well (waits 10 seconds and retries, technically infinite recursion if the error continues so uhh...might want to clean that one up at some point) but not heavily tested


tybug:
wrong branch I gotta figure out how to stop myself from doing that

PR #56, opened 2019-02-10
implement screening, closes #31

### LINKED ISSUE #31: add profile checker flag
in depth check of a user's profile, checking for both replay remodding and stealing on their top plays

samuelhklumpers:


tybug:
Pretty happy with this PR, screening still has a limitation of requiring the same number for num maps and num replays loaded per map, but that fix should come with a gui rewrite (gui needs some love! Add a row per map with a checkbox and a textbox each - whether to check and how many to check - when you enter a user's id perhaps)

@Smitty1298 @samuelhklumpers review please

PR #66, opened 2019-03-10
Fix against comparing replays of the same user

InvisibleSymbol:
sry


InvisibleSymbol:
Also I if remember correctly, this still uses two API calls for the user when he's in the lb. Maybe we can flip the order so it first downloads the lb and then checks if it still need to download the replay of the user.

tybug:
With the way we have it set up I think checking it before loading would be too messy. This way is ok for now

InvisibleSymbol:
Can we merge this? I see nothing wrong with it (for now at least) and it would be a good QoL improvement.

tybug:
I didn't merge originally because with 2.0 coming along this change will be obsolete anyway, and I didn't want to have another hotfix. I suppose there's really nothing wrong with a new release just for this though. Thanks for the bump

PR #68, opened 2019-04-16
fix bug(s) caused by non-osrs in replays/

samuelhklumpers:
This replaces the `f != .DS_STORE` check and prevents further bugs from arising when unexpected filetypes end up in replays/.

Initially found when Good Morning set the folder type to music, which creates a hidden system desktop.ini file, which would get processed as a osr, throwing a `ValueError: 91 is not a valid GameMode`.


tybug:
lgtm, we'll have to keep this in mind for 2.0 and add some sort of option to ignore this check in case someone wants to change/remove the extension for whatever reason

PR #86, opened 2019-08-20
v2.2.0 Release

tybug:
* Add relax cheat detection (and consequently UR calculation)
* Allow circleguard to be used without a database
* Add mods argument to map_check
* Retry requests if JSONDecodeError response is returned by ossapi
	* Avoids fatal error while replay loading if api returns invalid response
* Fix `pip install circleguard` failing if requirements were not installed
* Minor readme example updates
* Remove load progress tracking from Loader

## Now Using ossapi 1.2.2
* Return custom response for JSONDecodeError when api returns invalid json

## Now using circleparse 6.1.0
* replay_id now parses to an int instead of a tuple
* Add https://github.com/osufx/osu-parser files for more complicated parsing
* Switch license to GPL3 to comply with osu-parser license


tybug:
still needs config options for Detect. working on it now

PR #88, opened 2019-08-20
Detect mouse buttons in _parse_keys

InvisibleSymbol:


circleguard/investigator.py
```diff
@@ -50,17 +51,21 @@ def _parse_beatmap(self, beatmap):
...
         for keypress in replay:
-            if keypress[3] % 5 == 0 and keypress[3] != 0:  # 5=> key1, 10 => key2, 15 => both keys
-                if keypress[3] != last:
-                    if last != 15:  # ignore if the user held both buttons and let go of one
-                        keypresses.append(keypress[:3])  # t,x,y
-                    last = keypress[3]
-            else:
-                last = 0
+            if self._check_keys(keypress[3]):
+                    keypresses.append(keypress[:3])  # t,x,y
```
tybug: is removing k from the list necessary? far as I can see it's not (filter_keypresses just indexes), I suppose this is a tradeoff between the time it takes to iterate over the inner keypresses array and the memory cost of storing keys when we don't need them

InvisibleSymbol: No reason to keep it. The Hit object doesn't use it. I can add it tho, I can imagine it being used in the future.

tybug: just seems weird to remove it. I'd keep it, the memory footprint is tiny

tybug:
lgtm

PR #96, opened 2019-09-04
Performance Optimization 

InvisibleSymbol:
* switch to slider library for beatmap parsing
* store `Keys` enums as `int`s for better performance with bitwise operators


tybug:
I'm extremely happy with branch as is. Ready to remove draft status and merge @InvisibleSymbol ?

InvisibleSymbol:
yep, lgtm c:

PR #103, opened 2019-11-03
Aim Correction detection and Investigator rework

tybug:
Adds aim correction detection, based on small angles and distances between data points. No tests for this currently—will come soon with a tests rework hopefully, based on actual cheated and legit replays.

~~Returns a list of lists right now, should definitely make that an object.~~ returns a list of `Snap` now.

```python
from circleguard import *
cg = Circleguard("key")
m = Map(221777, num=1)

loose = Check(m, CorrectionDetect())
strict = Check(m, CorrectionDetect(min_distance=4))
for r in cg.run(loose):
    print(r.snaps)
for r in cg.run(strict):
    print(r.snaps)
```


tybug:
@samuelhklumpers would appreciate you doing a pass with your numpy magic. It's pretty fast right now but I'm sure it could be faster. Relevant code is in https://github.com/circleguard/circlecore/pull/103/files#diff-19a686d19443f3bf476faeb3b8bc0121R67-R139

tybug:
from @samuelhklumpers:

> in the rare case that den == 0 (and not [0]) not doing den!=0 gives a runtimewarning
> but I guess you could do away with it

Will leave it in then

InvisibleSymbol:
Any reason why everything in the investigator class is a staticmethod? We already pass the beatmap/replay to the class, why not use it? I mean I guess now you can directly access ur for example with `Investigator.ur(rp,bm)` but now the only real use for initializing the class is `.investigate()`... At that point why not also make that a staticmethod (Or a simple function)?

tybug:
Were we to make everything a static method, we wouldn't be able to do any preprocessing or optimizations like `np.array` in `init`. The Inheritance/what methods yield what also becomes muddied. `Comparer.compare` and `Investigate.investigate` are analogous, but putting that into some other method not tied to a class would make that less clear. We don't right now, but it would be trivial to make a parent class for both `Comparer` and `Investigator`, were there a need for it.

As for `ur` and `aim_correction` being static methods, I think you lose a lot of clarity if you don't pass arguments to the methods. The two methods are acting on fundamentally different information (a replay and beatmap, and just the beatmap) and it is not at all clear just from the function definition of `def aim_correction(self)` what is going on. `def aim_correction(replay_data)` is more clear.

This is also sort of a holdover, if you want to call it that, from `Comparer`: we have `_compare_two_replays` as a static method there (now that I look at it, it should really take replay_data and not replay, but that's a discussion for another time).

I think I've been harsher with my word choice in this reply than I meant to be; it's really a stylistic choice about clarity. It's true that we do lose something with static methods; we have to pass `max_angle` and `max_distance` instead of having them be available through `self`. But in my opinion the tradeoff is worth it.

PR #104, opened 2019-11-06
manually clean up temp dir

InvisibleSymbol:
Fixes `[WinError 32]`

### DIFF
--- a/circleguard/circleguard.py
+++ b/circleguard/circleguard.py
@@ -2,10 +2,9 @@ from pathlib import Path
 import sys
 import itertools
 import os
-import shutil
 from os.path import isfile, join
 import logging
-from tempfile import mkdtemp
+from tempfile import TemporaryDirectory
 
 from circleguard.loader import Loader
 from circleguard.comparer import Comparer
@@ -57,10 +56,11 @@ class Circleguard:
         self.loader = Loader(key, cacher=self.cacher) if loader is None else loader(key, self.cacher)
         if slider_dir is None:
             # have to keep a reference to it or the folder gets deleted and can't be walked by Library
-            self.__slider_dir = mkdtemp()
-            self.library = Library(self.__slider_dir)
+            self.__slider_dir = TemporaryDirectory()
+            self.library = Library(self.__slider_dir.name)
         else:
             self.library = Library(slider_dir)
+
 
     def run(self, check):
         """
@@ -135,6 +135,7 @@ class Circleguard:
         """
         info_loadable.load_info(self.loader)
 
+
     def set_options(self, cache=None):
         """
         Sets options for this instance of circlecore.
@@ -153,11 +154,6 @@ class Circleguard:
                 self.cache = cache
                 self.cacher.should_cache = cache
                 continue
-
-    def __del__(self):
-        self.library.close()
-        shutil.rmtree(self.__slider_dir)
-
 
 def set_options(loglevel=None):
     """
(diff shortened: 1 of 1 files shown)


InvisibleSymbol:
That confused me a bit, sorry that the response took so long.

First off, the Investigator has nothing to do with the Library (or at it's current implementation at least, unsure if that's something you want to change...)
The Circleguard class is the only thing that interacts with the slider library, so I guess we can... create a new library every time there is a relax check and deleted it after that? (Note that simply connecting & disconnecting doesn't resolve the issue at hand)
The problem is that we don't know _when_ the user is finished with processing everything, creating a fresh one every time is really bad for the api- If we choose to go with that it would be better to force a slider_dir and move the temp dir stuff onto the user...

Also, the worst thing that can happen with my code is that the dir won't be deleted, not really an important issue imo. (Can be tested by simply removing the `__del__` function; no exceptions, just some garbage left in the temp folder)

tybug:
I was implying that `Investigator` would have to be changed to deal with slider, yes.

> simply connecting and disconnecting doesn't resolve the issue at hand

Running the following code on my machine doesn't seem to agree with that:

```python
from tempfile import TemporaryDirectory
import sqlite3

class A():

    def __init__(self, close):
        self.temp_dir = TemporaryDirectory()
        open(self.temp_dir.name + "/db.db", 'a').close() # create db.db in temp dir
        self.db = sqlite3.connect(self.temp_dir.name + "/db.db")
        if close:
            self.db.close()
        
a = A(close=True) # no error
a = A(close=False) # error
```

We don't need to know when the user is done with everything because we wouldn't ever remove the dir until TempDir does it for us, just close the db which prevents the permissions error.

It's true that this is pretty much what the tmp folder is made for, and it's not too much of a concern if we leave garbage in there, but I'd like to do it right regardless.

InvisibleSymbol:
It seems like I approached this from the wrong direction, your way makes way more sense. I will rewrite this after #103 gets merged... Or, since it got merged while I was writing this, tommorow :+1:

circleguard/circleguard.py
```diff
@@ -56,8 +56,11 @@ def __init__(self, key, db_path=None, slider_dir=None, loader=None, cache=True):
         self.loader = Loader(key, cacher=self.cacher) if loader is None else loader(key, self.cacher)
         if slider_dir is None:
             # have to keep a reference to it or the folder gets deleted and can't be walked by Library
-            self.__slider_dir = TemporaryDirectory()
-            self.library = Library(self.__slider_dir.name)
+            self.slider_dir = TemporaryDirectory()
+            # create db and immediately disconnect
+            tmp_library = Library.create_db(self.slider_dir.name)
```
tybug: There's no advantage to using `create_db` here because there are no pre-existing osu files. We don't need to make a library at all here; rely on 

```python
if self.library is None:
    # connect to library since it's a temporary one
    library = Library(self.slider_dir.name)
```

later on to create the db the first time.

circleguard/circleguard.py
```diff
@@ -98,15 +101,28 @@ def run(self, check):
             comparer = Comparer(d.steal_thresh, compare1, replays2=compare2)
             yield from comparer.compare()
 
+        # relax check
         if Detect.RELAX in d or Detect.CORRECTION in d:
+            if Detect.RELAX in d:
+                if self.library is None:
```
tybug: `if not self.library`

don't directly compare to `None` unless you need to distinguish between `None` and falsey, which you don't here.

PR #111, opened 2019-11-21
v3.0.0 release

tybug:
* add aim correction detection
* new forward-facing documentation built with sphinx, including a comprehensive introduction on how to use circlecore
* new `User` class which represents a user's top plays
* new `MapUser` class which represents all of a user's plays on a map
* new `Mod` and `ModCombination` class which represent mods.
* restructure `replay.py` and inheritance of `loadables`. `Container` is (roughly) replaced with `InfoLoadable`, and `Check` is the only entry point for `Detect`.
* major `Detect` restructure; split into subclasses per cheat type. A `Detect` is instantiated with its respective thresholds(steal, ur, etc) instead of thresholds being global settings
* `Map` and `User` are now iterable and indexable, referencing `Replay`s in the `Map` or `User`
* rewrite and update all internal documentation
* switch documentation style from google to numpy (<https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard>)
* new method `cg.load_info` that loads the info for `ReplayContainer`s
* `Circleguard` now takes a `cache` argument, which can make the database effectively read-only
* global settings almost entirely removed, save for `loglevel`
* instance settings almost entirely removed, save for `cache`

- rename `replay.py` to `loadable.py`
- rename `UserInfo` to `ReplayInfo`
- `Options` class removed
- add a `loader#username` function which retrieves a username from a user id. See also https://github.com/ppy/osu-api/issues/281 
- add an `lru_cache` to `loader#map_id`, `#user_id`, and `#username` functions
- `loader#get_user_best` now returns a list of `UserInfo`, and accepts a `mods` argument.
- `cg.run` now only accepts a `Check`
- replace int mods with `ModCombination` in most places
- add ScoreV2 mod. Fixes not being able to process ScoreV2 replays.
- create a slider `Library` every time `#run` is called if `slider_dir` is not passed. Fixes `PermissionError`s on windows.
- remove convenience methods (`user_check`, `map_check`, etc). These have been replaced by `Map` and `User` (new)
- update STYLE document
- all Circleguard instances now use the same logger
- `filter` argument removed throughout the codebase
- `RatelimitWeight` and `ResultType` enum string values capitalized
- `ResultType.AIM_CORRECTION` renamed to `CORRECTION`
- `#ur` is now a staticmethod in `Investigator`


source/representing-replays.rst
```diff
@@ -0,0 +1,144 @@
...
+
+ReplayPath
+~~~~~~~~~~
+
+|ReplayPath| represents a replay stored locally in an ``osr`` file. For
+instance,
+
+.. code-block:: python
+
+    r2 = ReplayPath("/Users/tybug/Desktop/replays/replay1.osr")
```
InvisibleSymbol:
```suggestion
    r2 = ReplayPath("./tests/resources/example_replay.osr")
```

source/representing-replays.rst
```diff
@@ -0,0 +1,144 @@
...
+
+represents the replay in the file ``/Users/tybug/Desktop/replays/replay1.osr``.
```
InvisibleSymbol:
```suggestion
represents the replay in the file ``./tests/resources/example_replay.osr``.
```

source/representing-replays.rst
tybug: This code isn't meant to be run in a circlecore directory; I'm just giving an example path. Maybe `"/path/to/your/replay.osr` is better, but I think a real example of a path works too.

Using the test osrs is just confusing.

InvisibleSymbol: yea, `"/path/to/your/replay.osr"` works way better, `"/Users/tybug/Desktop/replays/replay1.osr"` just feels unpolished imo.

PR #116, opened 2019-12-05
sphinx version and footer fix

tybug:
Cleaner footer, use `__version__` from circleguard. I don't think this will actually change anything about our docs until we self host and move off rtd.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -4,9 +4,9 @@
 # list see the documentation:
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 
-# This will fail if circlecore's dependencies aren't installed.
-# Which shouldn't be an issue because the only people running ``make html``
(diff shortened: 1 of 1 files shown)


tybug:
This isn't necessary and is actually rather harmful. `from circleguard import __version__` already uses your local changes if you installed it editably (that is, `pip install -e /path/to/circlecore/dir/`).

InvisibleSymbol:
Well, my change was primarily _so you don't have to install it_, since I was cautious about somebody forgetting to replace their global pypi version. If you use an venv or just not use pypi it probably won't be an issue, though if somebody just wants to work on the docs (and test them), the "force local files" way makes more sense imo.

tybug:
Installing locally or through pypi is the proper way to use packages. We should not be encouraging stupidity and roundabout methods when the official way works perfectly well.

PR #118, opened 2019-12-07
Add Macro detection

InvisibleSymbol:
Basically looks at how many extremely short presses were done in a replay. If it's over a threshold, it's cheated.


circleguard/enums.py
```diff
@@ -292,6 +293,8 @@ def __init__(self, value):
         self.ur_thresh = None
```
tybug: t

tybug:
Got sidetracked by mildly rewriting `Detect` to allow for statements like `d.requires(Detect.REQUIRES_BEATMAP)` or `Detect.INVESTIGATION in d`, which will reduce nasty multiple if statements. That should be a separate pr but I'm lazy and will push it here. Will review/complete that tomorrow

tybug:
this branch now returns a wildly different ur than release:

```python
rm = Map(221777, num=1)
print(list(cg.run(Check(rm, RelaxDetect())))[0].ur)
```

```
>>> 64.50578685582167
>>> 1901.8454178099878
```

I suspect it's because you're using `hit.press.press_length` in `ur`. Should be `hit_error`?

tybug:
Still not accurate ur-wise for toy's play on 221777. osu client (and release) gives 64.51 (couldn't take a ss because of issues with fullscreen. Feel free to watch it yourself and confirm), HEAD here gives 64.61. It seems to pass the test cases fine, so there's some edge case where behavior changed here.

tybug:
Running line profiler shows more than a 2x increase in time per ur calculation:

`python3 -m line_profiler relax_release_branch.lprof`
```
Timer unit: 1e-06 s

Total time: 0.299924 s
File: /Users/tybug/Desktop/coding/sandbox/_randompython.py
Function: run_relax_check at line 13

    13                                           @profile
    14                                           def run_relax_check():
    15        10         25.0      2.5      0.0      d = cg.run(c)
    16        10     299899.0  29989.9    100.0      list(d)
```

`python3 -m line_profiler relax_macro_branch.lprof`
```
Timer unit: 1e-06 s

Total time: 0.713971 s
File: /Users/tybug/Desktop/coding/sandbox/_randompython.py
Function: run_relax_check at line 13

    13                                           @profile
    14                                           def run_relax_check():
    15        10         27.0      2.7      0.0      d = cg.run(c)
    16        10     713944.0  71394.4    100.0      list(d)
```

Test code:


```python
cg = Circleguard("key", db_path="./db.db")
r3 = ReplayPath("/Users/tybug/Desktop/coding/osu/circlecore/tests/resources/legit_replay1.osr")
rm = Map(221777, num=1)
c = Check(r3, RelaxDetect())
cg.load(c)

@profile
def run_relax_check():
    d = cg.run(c)
    list(d)

for i in range(10):
    run_relax_check()
```

Don't have time to investigate further right now, but this PR can't go forward without equaling or increasing performance relative to release.

tybug:
Going to close mostly per https://github.com/circleguard/circlecore/pull/119#issuecomment-796606066. The only thing new in this branch is seeing how many keypresses in a replay are under a certain number of ms. Something we should have, but that would need rewriting anyway given how much circlecore has changed.

PR #119, opened 2019-12-07
Cleaning tests

samuelhklumpers:
Rewrote some of Comparer and added some methods in an attempt to circumvent some effects that may cause false negatives when comparing, while also slightly widening the gap in simlarity values between legit comparisons and copied ones.

Fails the hardcoded similarity value tests:
```
======================================================================
FAIL: test_cheated_replaypath (test.TestReplays)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\Samuel Klumpers\Desktop\Programming\Python\circleguard\src\circleguard\tests\test.py", line 53, in test_cheated_replaypath
    self.assertAlmostEqual(r.similarity, 4.2608, delta=0.0001, msg="Similarity is not correct")
AssertionError: 1.8745389167639144 != 4.2608 within 0.0001 delta (2.3862610832360853 difference) : Similarity is not correct

======================================================================
FAIL: test_legitimate_replaypath (test.TestReplays)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\Samuel Klumpers\Desktop\Programming\Python\circleguard\src\circleguard\tests\test.py", line 75, in test_legitimate_replaypath
    self.assertAlmostEqual(r.similarity, 24.2129, delta=0.0001, msg="Similarity is not correct")
AssertionError: 23.17843859264939 != 24.2129 within 0.0001 delta (1.0344614073506122 difference) : Similarity is not correct

----------------------------------------------------------------------
```

Some tests comparing the previous methods, the combination of alignment/synchronization/validation and searching/validation. Using Woey's replay, a cheated copy, and three shifted copies on https://osu.ppy.sh/beatmapsets/773330#osu/1626537 : 
```
clean mode                               :  none,   all, searc
cheater - Woey                           :  1.22,  0.80,  0.73
cheater - xy_shift                       :  1.41,  0.01,  0.00
cheater - t_shift                        :  9.54,  9.56,  0.00
cheater - txy_shift                      :  9.87,  9.56,  0.00
Woey - xy_shift                          :  1.98,  0.80,  0.72
Woey - t_shift                           :  9.49,  9.52,  0.72
Woey - txy_shift                         :  9.82,  9.52,  0.72
xy_shift - t_shift                       :  9.42,  9.55,  0.00
xy_shift - txy_shift                     :  9.54,  9.55,  0.00
t_shift - txy_shift                      :  1.41,  0.00,  0.00
```


circleguard/comparer.py
```diff
@@ -103,15 +108,69 @@ def _result(self, replay1, replay2):
             The result of comparing ``replay1`` to ``replay2``.
         """
         self.log.log(utils.TRACE, "comparing %r and %r", replay1, replay2)
-        result = Comparer._compare_two_replays(replay1, replay2)
-        mean = result[0]
-        sigma = result[1]
+
+        if CleanMode.SEARCH in self.detect.clean_mode:
+            mean = Comparer._compare_hill_climb(replay1, replay2, self.detect.clean_mode)
+        else:
```
tybug: It seems then that there are really two distinct options in `CleanMode` - preprocessing (`VALIDATE`, `SYNCHRONIZE`, `ALIGN`) and processing (actual comparison) (`SEARCH`). Unless I'm misunderstanding that, using a single variable - `clean_mode` - for both of them is very confusing.

circleguard/enums.py
```diff
@@ -369,6 +375,45 @@ def __init__(self, max_angle=10, min_distance=8):
...
+        The time interval used when searching.
+    step_limit: int
+        The maximal amount of steps performed when searching.
+    """
+
+    VALIDATE = 1 << 0
+    SYNCHRONIZE = 1 << 1
+    ALIGN = 1 << 2
+    SEARCH = 1 << 3
+    ALL = VALIDATE + SYNCHRONIZE + ALIGN + SEARCH
```
tybug: 1-2 sentence comment about each of these would be nice. I know there's a lengthier explanation in each of their methods (`filter_single`, `align_clocks`, `align_coordinates`, etc) but I don't want to have to hunt those methods down to know at a glance what an option does.

tybug:
Configuring your linter in whatever IDE you use to remove whitespaces at the end of a line would be nice (avoids commits like 26f0e327244d4255191019d0401e0682b0adf133, which are both annoying and clutter git history).

circleguard/loadable.py
```diff
@@ -706,3 +706,317 @@ def load(self, loader, cache):
         Replay.__init__(self, loaded.timestamp, map_id, loaded.player_name, user_id, ModCombination(loaded.mod_combination),
                         loaded.replay_id, loaded.play_data, self.weight)
         self.log.log(TRACE, "Finished loading %s", self)
+
+
+class ReplayModified(Replay):
```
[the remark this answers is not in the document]
samuelhklumpers: Modifying the `Replay` using the new methods only makes sense in the context of comparison,
so it may be more suitable to use `copy.deepcopy` and move these methods to a helper class or `Comparer`

circleguard/loadable.py
```diff
@@ -706,3 +706,317 @@ def load(self, loader, cache):
...
+            The copy as a :class:`~.ReplayModified`.
+        """
+        
+        r = ReplayModified(replay.timestamp, replay.map_id, replay.username,
+                      replay.user_id, replay.mods, replay.replay_id, replay.replay_data, replay.weight)
+        r.txyk = replay.as_list_with_timestamps()
+        return r
+
+    @staticmethod
+    def filter_single(replay):
```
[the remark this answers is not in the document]
samuelhklumpers: I think that only `filter_single` is useful outside of comparison so that could be moved to `utils` or `Replay` but the others should be kept close to the comparison

circleguard/comparer.py
samuelhklumpers: There are cases in which invalid data is stolen so using only `SYNCHRONIZE + ALIGN` is wanted, some framerate mismatches may cause `SYNCHRONIZE` to behave weirdly and only interpolating between two replays is preferable. I do agree that `VALIDATE + SEARCH` could be mutually exclusive to `SYNCHRONIZE + ALIGN`.

circleguard/loadable.py
```diff
@@ -542,6 +544,24 @@ def __repr__(self):
     def __str__(self):
         return f"Replay by {self.username} on {self.map_id}"
 
+    def as_arrays(self):
```
tybug: would prefer this logic to happen in `__init__`; it's 6 extra lines but nothing too complicated. And the logic only happens once so we shouldn't be making it available as a function.

circleguard/loadable.py
```diff
@@ -542,6 +544,24 @@ def __repr__(self):
...
+        block = list(zip(*[(e.time_since_previous_action, e.x, e.y, e.keys_pressed) for e in self.replay_data]))
+
+        t = np.array(block[0], dtype=int)
+        t = t.cumsum()
+        xy = np.array([block[1], block[2]], dtype=float).T
+        k = np.array(block[3], dtype=int)
+
+        return t, xy, k
+
     def as_list_with_timestamps(self):
```
tybug: Can this method be removed in favor of the new `t`, `xy`, `k`? Only `Investigator` uses it right now but I'm pretty sure it can be changed to use `ndarrays` instead of this method.

circleguard/enums.py
```diff
@@ -321,10 +323,14 @@ class StealDetect(Detect):
     steal_thresh: float
         If the average distance in pixels of two replays is smaller than
         this value, they are labeled cheated (stolen replays). Default 18.
+    clean_mode: :class:`~.CleanMode`
+        The options used to clean the replays before comparing them.
     """
-    def __init__(self, steal_thresh=18):
+    def __init__(self, steal_thresh=18, clean_mode=None):
```
tybug: We should pick a sensible default for clean_mode - `CleanMode.SLOW` perhaps? Instead of making the user choose.

tybug:
Some disturbing results from `line_profiler`:

```
Timer unit: 1e-06 s

Total time: 16.314 s
File: /Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py
Function: compare at line 57

    57                                               @profile
    58                                               def compare(self):
    59                                                   """
    60                                                   If ``replays2`` is not ``None``, compares all replays in replays1
    61                                                   against all replays in replays2. Otherwise, compares all replays in
    62                                                   ``replays1`` against all other replays in ``replays1``
    63                                                   (``len(replays1) choose 2`` comparisons).
    64                                           
    65                                                   Yields
    66                                                   ------
    67                                                   :class:`~.result.ComparisonResult`
    68                                                       Results representing the comparison of two replays.
    69                                                   """
    70                                           
    71         1         14.0     14.0      0.0          self.log.info("Comparing replays with mode: %s", self.mode)
    72         1          4.0      4.0      0.0          self.log.debug("replays1: %r", self.replays1)
    73         1          3.0      3.0      0.0          self.log.debug("replays2: %r", self.replays2)
    74                                           
    75                                                   #TODO: a little bit hacky and I don't think works 100% correctly, if mode is double but replays2 is None
    76         1          1.0      1.0      0.0          if not self.replays1 or self.replays2 == []:
    77                                                       return
    78                                           
    79         1          2.0      2.0      0.0          if self.mode == "single" and self.detect.clean_mode.value:
    80         1   16214163.0 16214163.0     99.4              self.replays1 = Comparer.clean_set(self.replays1, self.detect.clean_mode)
    81                                           
    82         1          2.0      2.0      0.0          if self.mode == "double":
    83                                                       iterator = itertools.product(self.replays1, self.replays2)
    84         1          1.0      1.0      0.0          elif self.mode == "single":
    85         1         11.0     11.0      0.0              iterator = itertools.combinations(self.replays1, 2)
    86                                                   else:
    87                                                       raise InvalidArgumentsException("'mode' must be one of 'double' or 'single'")
    88                                           
    89        46         71.0      1.5      0.0          for replay1, replay2 in iterator:
    90        45         84.0      1.9      0.0              if replay1.replay_id == replay2.replay_id:
    91                                                           self.log.debug("Not comparing %r and %r with the same id", replay1, replay2)
    92                                                           continue
    93        45      99658.0   2214.6      0.6              yield self._result(replay1, replay2)

Total time: 16.2141 s
File: /Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py
Function: clean_set at line 467

   467                                               @staticmethod
   468                                               @profile
   469                                               def clean_set(replays, mode):
   470                                                   """
   471                                                   Cleans the :class:`~.Replay`s in replays using the methods specified in the mode.
   472                                           
   473                                                   Parameters
   474                                                   ----------
   475                                                   replays: list(:class:`~.Replay`)
   476                                                       The replays to clean.
   477                                                   mode: :class:`~.CleanMode`
   478                                                       The mode specifying the used methods.
   479                                           
   480                                                   Returns
   481                                                   -------
   482                                                   list(:class:`~.Replay`)
   483                                                       The cleaned :class:`~.Replay`s
   484                                                   """
   485                                           
   486         1   16185719.0 16185719.0     99.8          replays = [copy.deepcopy(r) for r in replays]
   487                                           
   488         1         34.0     34.0      0.0          if CleanMode.VALIDATE in mode:
   489         1      15628.0  15628.0      0.1              replays = [Comparer.filter(replay) for replay in replays]
   490                                           
   491         1          7.0      7.0      0.0          if CleanMode.SYNCHRONIZE in mode:
   492                                                       timestamps = [replay.t for replay in replays]
   493                                           
   494                                                       timestamps = Comparer.align_clocks(timestamps)
   495                                           
   496                                                       replays = [Comparer.interpolate_to(replay, timestamps) for replay in replays]
   497                                           
   498         1          2.0      2.0      0.0          if CleanMode.ALIGN in mode:
   499         1      12695.0  12695.0      0.1              replays = Comparer.align_coordinates(replays)
   500                                           
   501         1          2.0      2.0      0.0          return replays

Total time: 16.3448 s
File: /Users/tybug/Desktop/coding/sandbox/_randompython.py
Function: compare_replays at line 18

    18                                           @profile
    19                                           def compare_replays():
    20         2        476.0    238.0      0.0      for _ in range(ITERATIONS):
    21        46   16342198.0 355265.2    100.0          for r in cg.run(c):
    22        45       2134.0     47.4      0.0              print(r.similarity)
```

`deepcopy` is taking up a ridiculous amount of time, and frankly was probably never the right way to go about this. Can we not deal in `ndarrays` instead of `replay` objects? ie, have `Comparer.filter`, `Comparer.interpolate_to`, and `Comparer.align_coordinates` accept some variation of `t`, `xy`, and `k` `ndarrays` instead of the entire `replay`, just like `Comparer.align_clocks` does.

tybug:
48c7081c2153cc349dccd7e98f12c881d6a3eeb6 errors for me on steal detection:

```
Traceback (most recent call last):
  File "/Users/tybug/Desktop/coding/_RESOURCES/line_profiler/kernprof.py", line 234, in <module>
    sys.exit(main(sys.argv))
  File "/Users/tybug/Desktop/coding/_RESOURCES/line_profiler/kernprof.py", line 222, in main
    execfile(script_file, ns, ns)
  File "/Users/tybug/Desktop/coding/_RESOURCES/line_profiler/kernprof.py", line 35, in execfile
    exec_(compile(f.read(), filename, 'exec'), globals, locals)
  File "/Users/tybug/Desktop/coding/sandbox/_randompython.py", line 27, in <module>
    compare_replays()
  File "/Users/tybug/Desktop/coding/_RESOURCES/line_profiler/line_profiler.py", line 119, in wrapper
    result = func(*args, **kwds)
  File "/Users/tybug/Desktop/coding/sandbox/_randompython.py", line 23, in compare_replays
    for r in cg.run(c):
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/circleguard.py", line 100, in run
    yield from comparer.compare()
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py", line 79, in compare
    Comparer.clean_set(self.replays1, self.detect.clean_mode)
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py", line 438, in clean_set
    for i, xy in enumerate(Comparer.align_coordinates([d[1] for d in data])):
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py", line 401, in align_coordinates
    m = xys[0].mean(axis=0)
AttributeError: 'list' object has no attribute 'mean'
```

A couple print statements shows that `align_coordinates` is being passed a list of lists, not ndarrays:
```
        print(type(xys))
        print(type(xys[0]))
```

```
<class 'list'>
<class 'list'>
```

samuelhklumpers:
I see, I only tested offline for which it doesn't error, good thing you're testing 👍

tybug:
Interesting results when run on the first 50 replays from https://osu.ppy.sh/beatmapsets/79498#osu/221777, with all possible combinations of `CleanMode`. In https://cdn.discordapp.com/attachments/533402800390930435/663579410854838282/cleaning_mode_test_results.txt, the top row represents the `CleanMode` used, in the order `SEARCH, ALIGN, SYNCHRONIZE, VALIDATE` (so `0101` used `ALIGN` and `VALIDATE`) and the leftmost column is the two replays being compared. The matrix between them is the similarity for that `CleanMode` and set of replays. 

`1011` and `1101` appear to have the same results, but not the same as `1111`, implying that they act identically separately but differently when combined?

And thankfully, none of the values are wildly different. The greatest difference in a row is 5 or so. Unsurprisingly, `0001` (just `VALIDATE`) consistently has the highest similarly in a row. This is all tested on presumably clean data, though, so I think it would be good to keep `VALIDATE` for when we run into replays that do have out of bounds data that would otherwise throw us off.

I profiled this same run as well; `lprof` file [here](https://cdn.discordapp.com/attachments/533402800390930435/663579414474391583/cleaning_mode_performance.lprof). I was very surprised to see `ALIGN` at just `0.751043` seconds while `VALIDATE` was slower, at `1.807` seconds. I trust the general scale of the profile but this was only profiling a single iteration so I want to run `ALIGN` and `VALIDATE` both again with `ITERATIONS=10` and see if `ALIGN` is really faster for some strange reason.

Speaking of the general scale, this is extremely fast; an improvement over the master branch for sure, especially when using `CleanMode.FAST` or something similar. The longest of the combinations took around 25 seconds for 50 replays on this 6 minute map. I don't remember exactly how well master would have done, but I think a lot worse than that.

tybug:
I'm very worried by the similarity results changing when the size of the replay array differs. ie, ideally, the comparison of the `#1` and `#2` play for `m = Map(221777, NUM)` should return the same regardless of if `NUM` is `2` or `50`. That's currently not the case for any `CMode` except for `Align`. I wouldn't be so worried if things worked perfectly, but of course they won't - we'll need to debug and asking people exactly what combination of replays they used when comparing, as well as people getting different similarity results from each other because of seemingly irrelevant variables (number of other replays in the `ReplayContainer`), isn't ideal.

I ran this branch through 7 sets of replay steals, and with a low (0 or 1) number of additional replays it works great. But anything after (and including) `0011` gives very large similarity values for stolen replays (50 vs 5). Here's the txt results of that: [cmode_2+14steal_all.txt](https://github.com/circleguard/circlecore/files/4144983/cmode_2%2B14steal_all.txt)

and the script (slightly changed from previous): https://drive.google.com/file/d/1rCzByvNUyu7GEeFvoye63ghuQjaTE5LD/view?usp=sharing

and the folder the replays come from: https://drive.google.com/drive/folders/1Mif1Z3NT5SnLa0NROkHMa5Beg7JWLuaF?usp=sharing

tybug:
It's been over a year since this pr (and 2 years since circlecore started) and I don't think we've seen anything that would warrant deeper cleaning like this branch has. If we ever happen to need it we can look at this pr for reference, but the circlecore api has changed so much this code would need rewriting anyway.

PR #122, opened 2019-12-23
Fix loader

InvisibleSymbol:
`response` wasn't being checked, array operation caused an `KeyError` when an invalid key was specified. Now raises an correct `InvalidKeyException`.

### DIFF
--- a/circleguard/loadable.py
+++ b/circleguard/loadable.py
@@ -663,6 +663,7 @@ class ReplayPath(Replay):
     def __init__(self, path, cache=None):
         self.log = logging.getLogger(__name__ + ".ReplayPath")
         self.path = path
+        self.hash = None
         self.cache = cache
         self.weight = RatelimitWeight.LIGHT
         self.loaded = False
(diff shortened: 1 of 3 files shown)


circleguard/loadable.py
```diff
@@ -663,6 +663,7 @@ class ReplayPath(Replay):
     def __init__(self, path, cache=None):
         self.log = logging.getLogger(__name__ + ".ReplayPath")
         self.path = path
+        self.hash = None
```
tybug: is this hash being used for anything?

tests/test.py
```diff
@@ -203,6 +204,47 @@ def test_map_user_slice(self):
         # test slicing
         self.assertListEqual([r.map_id for r in self.mu[0:2]], [795627, 795627])
 
+
+class TestLoader(CGTestCase):
```
tybug: combine the methods of this class; merge `test_incorrect_map_id` with `test_working_map_id` (to a method named `test_loading_map_id`, same with the three `user_id` methods and the two `username` methods.

tests/test.py
```diff
@@ -203,6 +204,47 @@ def test_map_user_slice(self):
...
+    def test_incorrect_username(self):
+        result = self.loader.username(0)
+        self.assertEqual(result, "")
+
+    def test_working_username(self):
+        result = self.loader.username(13506780)
+        self.assertEqual(result, "] [")
+
+    def test_incorrect_key(self):
+        loader = Loader("LULW 727 LULW")
```
tybug: use a more professional key please :p `"incorrect key"` works

circleguard/loadable.py
InvisibleSymbol: If the api fails this is the only way to Identify a map. Same as how the username is kept even if we get the user id.

PR #125, opened 2020-02-01
allow Mod to be instantiated, add string instantiation

tybug:
* mods can now be instantiated through `Mod`. Instantiation through `ModCombination` is highly discouraged.
* mods can now be instantiated with a string, not only an int. The string
must be a combination, in any order, of the two length strings that represent mods. For instance, `Mod("HDHR")` is valid, and `Mod("HDHR") == Mod.HDHR` is `True`.

### DIFF
--- a/circleguard/enums.py
+++ b/circleguard/enums.py
@@ -273,9 +273,9 @@ class Mod(ModCombination):
              FI, RD, CN ,TP, K1, K2, K3, K4, K5, K6, K7, K8, K9, CO, MR]
 
     def __init__(self, value):
-        if type(value) is int:
+        if isinstance(value, int):
             super().__init__(value)
-        if type(value) is str:
+        if isinstance(value, str):
             super().__init__(Mod._parse_mod_string(value))
 
     @staticmethod
@@ -309,7 +309,7 @@ class Mod(ModCombination):
             single_mod_string = mod_string[i - 2: i]
             # there better only be one Mod that has an acronym matching ours,
             # but a comp + 0 index works too
-            matching_mods = [mod for mod in Mod.ALL if mod.short_name() == single_mod_string]
+            matching_mods = [mod for mod in Mod.ORDER if mod.short_name() == single_mod_string]
             if not matching_mods:
                 raise ValueError(f"Invalid mod string (no matching mod found for {single_mod_string})")
             mod_value += matching_mods[0].value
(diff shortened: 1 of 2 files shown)


circleguard/enums.py
```diff
@@ -267,6 +272,52 @@ class Mod():
              V2, TD, # we stop caring about order after this point
              FI, RD, CN ,TP, K1, K2, K3, K4, K5, K6, K7, K8, K9, CO, MR]
 
+    # same as ORDER but with NM included
+    ALL = [NM, EZ, HD, HT, DT, _NC, HR, FL, NF, SD, _PF, RX, AP, SO, AT,
+           V2, TD, FI, RD, CN ,TP, K1, K2, K3, K4, K5, K6, K7, K8, K9, CO, MR]
```
InvisibleSymbol: These two array can be combined.
No reason to not have NM in order as far as I can see

circleguard/enums.py
```diff
@@ -267,6 +272,52 @@ class Mod():
...
+
+    def __init__(self, value):
+        if type(value) is int:
```
InvisibleSymbol: wouldn't be `isinstance(value, int)` be better?

circleguard/enums.py
```diff
@@ -267,6 +272,52 @@ class Mod():
...
+            If mod_string is empty, not of even length, or any of its 2-length
+            substrings do not correspond to a ModCombination in Mod.ALL.
+        """
+        if mod_string == "":
+            raise ValueError("Invalid mod string (cannot be empty)")
+        if len(mod_string) % 2 != 0:
+            raise ValueError(f"Invalid mod string {mod_string} (not of even length)")
+        mod_value = 0
+        for i in range(2, len(mod_string) + 1, 2):
+            single_mod_string = mod_string[i - 2: i]
```
InvisibleSymbol:
```suggestion
        for i in range(2, len(mod_string) - 2, 2):
            single_mod_string = mod_string[i: i + 2]
```
easier to understand imo

circleguard/enums.py
tybug: hm, I was thinking that `NM` wouldn't make much sense in `ORDER` because nobody writes +NM in scorepost titles, but it wouldn't break anything to change it and makes sense coding wise

circleguard/enums.py
tybug: yes, meant to change it to that (though in practice it doesn't matter since nobody really subclasses primitives)

PR #132, opened 2020-03-29
Add hill climbing test and CMode structure

tybug:
* adds some semblance of the `CleanMode` structure the `cleaning-tests` branch had, but renamed to `CMode` (comparison mode), as I felt that was more descriptive.
* `SlowCMode` now tests both hill climbing and our standard comparison algorithm, returning the lower similarity of the two. Names and documentation have been updated to reflect this, I think
* should allow for users (read: circleguard) to transfer over to using hill climbing for searching easily, if they so choose

### DIFF
--- a/circleguard/comparer.py
+++ b/circleguard/comparer.py
@@ -105,6 +105,8 @@ class Comparer:
         """
         self.log.log(utils.TRACE, "comparing %r and %r", replay1, replay2)
         result = Comparer._compare_two_replays(replay1, replay2)
+        result2 = Comparer._compare_hill_climb(replay1, replay2)
+        print(f"normal comparison: {result[0]:.6f}, hill climb: {result2:.6f}")
         mean = result[0]
         sigma = result[1]
         ischeat = False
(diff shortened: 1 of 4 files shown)


circleguard/comparer.py
```diff
@@ -154,8 +163,83 @@ def _compare_two_replays(replay1, replay2):
...
+          Repeat the process, taking those new time values as the starting
+          point.
+        * If neither the shift to the left nor right produces a lower similarity
+          than the previous time values, we are at a local minimum. Return the
+          current similarity.
+        """
+
+        t1, xy1, k1 = replay1.t, replay1.xy, replay1.k
+        t2, xy2, k2 = replay2.t, replay2.xy, replay2.k
+        prev_value = math.inf # arbitrarily high value
```
InvisibleSymbol: shouldn't this be the mean at 0 shift? would prevent it from shifting if the current one is already the lowest.

circleguard/comparer.py
```diff
@@ -154,8 +163,83 @@ def _compare_two_replays(replay1, replay2):
...
+            # try shifting both backwards and forwards in time
+            for sgn in [-1, 1]:
+                replay1.t = t1 + sgn * dt
+                v = replay1.t
+
+                value = Comparer._compare_two_replays(replay1, replay2)
+
+                values[value] = v
+                replay1.t, replay1.xy, replay1.k = t1, xy1, k1
+                replay2.t, replay2.xy, replay2.k = t2, xy2, k2
```
InvisibleSymbol: why? `_compare_two_replays` doesn't seem to interact with the replays directly. I understand that you need to set `replay1.t` but everything else seems unnecessary

tybug:
Closing per https://github.com/circleguard/circlecore/pull/119#issuecomment-796606066. The only thing potentially worthwhile here is hill climbing, since that can detect time shifts, but #139 does a, I believe, better (although likely slower) job of that. We can revisit this branch if necessary in the future, but it will take a different form than this pr if at all.

PR #134, opened 2020-04-09
use preprocessed t and xy instead of as_list_with_timestamps

tybug:
should be identical (though I haven't tested, we need to add a test case for aim correction). This should also be done for ur calculation so we can finally be rid of `as_list_with_timestamps`.


tests/test.py
```diff
@@ -133,6 +134,30 @@ def test_num_invariance(self):
...
+        cls.r1 = ReplayPath(RES / "corrected_replay1.osr")
+
+    def test_cheated_replay(self):
+        r = list(self.cg.run(Check([self.r1], CorrectionDetect())))[0]
+        snaps = r.snaps
+
+        self.assertEqual(len(snaps), 15)
+        # beginning
+        self.assertEqual(snaps[0].time, 5103)
+        self.assertEqual(snaps[0].angle, 7.384911923724048)
```
InvisibleSymbol, line 150: how about using `assertAlmostEqual`?

tybug: yep

tests/test.py
```diff
@@ -2,8 +2,9 @@
 from unittest import TestCase, skip, TestSuite, TextTestRunner
 from pathlib import Path
 import warnings
-from circleguard import (Circleguard, Check, ReplayMap, ReplayPath, RelaxDetect, StealDetect,
+from circleguard import (Circleguard, Check, ReplayMap, ReplayPath, RelaxDetect, StealDetect, CorrectionDetect,
                          RatelimitWeight, set_options, Map, User, MapUser, Mod, Loader, InvalidKeyException)
+from circleguard.investigator import Snap
```
InvisibleSymbol, line 7: unused import

tybug: oh right

PR #135, opened 2020-04-13
rework ur calculation

tybug:
* use replay `t`, `xy`, `k` in ur calculation
* remove `replay#as_list_with_timestamps`


tybug:
we don't actually have ur tests atm, I tested it on one replay locally and it gave the same ur

PR #137, opened 2020-04-18
Cv ur

tybug:
makes `result.ur` cvur. Still stores ucv ur as a separate object


circleguard/result.py
```diff
@@ -106,14 +107,16 @@ class RelaxResult(InvestigationResult):
     replay: :class:`~circleguard.loadable.Replay`
         The replay investigated.
     ur: int
-        The unstable rate of the replay. More information on UR available at
-        https://osu.ppy.sh/help/wiki/Accuracy#accuracy
+        The (unconverted) unstable rate of the replay. More information on UR
+        available athttps://osu.ppy.sh/help/wiki/Accuracy#accuracy
```
InvisibleSymbol, line 111:
```suggestion
        available at https://osu.ppy.sh/help/wiki/Accuracy#accuracy
```

circleguard/utils.py
```diff
@@ -110,3 +110,24 @@ def span_to_list(span):
         else:
             ret.add(int(s))
     return ret
+
+def ur_to_cv_ur(ur, mods):
```
InvisibleSymbol, line 114: how about `convert_ur()` ?

circleguard/result.py
```diff
@@ -106,14 +107,16 @@ class RelaxResult(InvestigationResult):
...
     ischeat: bool
         Whether the replay is cheated or not.
     """
     def __init__(self, replay: Replay, ur: int, ischeat: bool):
         super().__init__(replay, ischeat, ResultType.RELAX)
-        self.ur = ur
+        self.ur = ur_to_cv_ur(ur, replay.mods)
```
InvisibleSymbol, line 117: imo `self.ur` should be what osu! returns, which is unconverted.

tybug: I had that originally and then switched it, reasoning being we're comparing against cv ur to determine `ischeat`, so we should return cv ur as `ur` and keep ucv ur as the odd one out. 

It would be confusing to have `result.ur` return something higher than your threshold, and `result.ischeat` return `True`.

circleguard/utils.py
tybug: that implies both directions, but that's actually good, I'll add an extra `to` argument which is either `"cv"` or `"ucv"` and make it `convert_ur`

circleguard/result.py
InvisibleSymbol: ok makes sense. how about renaming `ucv_ur` to either `raw_ur` or `uncv_ur`?

tybug: There doesn't seem to be a standard naming convention for unconverted ur (which means we get to make it up ourselves!) - I like `ucv_ur`, `raw` doesn't fit with the existing scheme (`cvUR`) and prepending `un` instead of `u` feels too vernose

circleguard/utils.py
InvisibleSymbol: I feel like a string as an argument seems overly complex. I would actually prefer the current way over that´.

tybug: unless we want to create a `ur` object (not necessarily a terrible idea, but an over engineered solution atm) or have a `ConversionMode` enum, a string argument is the best solution to providing both directions

InvisibleSymbol: ye but I mean `ur_to_cv_ur` and `cv_ur_to_ur` would be simpler than a string argument imo.

tybug: the code duplication there is so nasty :/ disagree

circleguard/utils.py
```diff
@@ -110,3 +110,27 @@ def span_to_list(span):
         else:
             ret.add(int(s))
     return ret
+
+def convert_ur(ur, mods, *, to):
```
InvisibleSymbol, line 114: missing doc for `to`. also `to` should have a default assigned imo

circleguard/utils.py
```diff
@@ -110,3 +110,27 @@ def span_to_list(span):
...
+    if to not in ["cv", "ucv"]:
+        raise ValueError(f"Expected one of cv, ucv. Got {to}")
+    conversion_factor = 1
+
+    if Mod.DT in mods:
+        conversion_factor = (1 / 1.5) if to == "cv" else 1.5
+    elif Mod.HT in mods:
+        conversion_factor = (1 / 0.75) if to == "cv" else 0.75
+
+    return ur * conversion_factor
```
InvisibleSymbol, line 136:
```suggestion
    if Mod.DT in mods:
        conversion_factor = 1.5
    elif Mod.HT in mods:
        conversion_factor = 0.75
    if to == "ucv":
        return ur * conversion_factor
    return ur / conversion_factor

```
This is way cleaner imo

tybug: you got them backwards, but agree

circleguard/utils.py
tybug: disagree on the default, `convert_ur` is already a direction-neutral statement, having a call of `convert_ur(10)` without the `to` argument is hard to understand

InvisibleSymbol: ok makes sense, but why is it a forced keyword only argument then? `convert_ur(10, mods, "cv")` looks fine to me

tybug: ambiguous between `to` or `from` (am I starting as cv or ending as cv?)

circleguard/result.py
```diff
@@ -1,5 +1,6 @@
 from circleguard.loadable import Replay
-from circleguard.enums import ResultType
+from circleguard.enums import ResultType, Mod
```
InvisibleSymbol, line 2: Why is `Mod` being imported? Doesn't look like it's being used in any changes you made

tybug: I..don't know how that got here

tybug: oh I do actually, I originally wrote the conversion in the result class

Issue #115, opened 2019-11-30
favor span over num in ReplayContainers


tybug:
Currently it's very easy to forget to pass one of `span` or `num` when creating a `ReplayContainer`:

```python
m = Map(221777)
```

since both arguments are marked optional, this is valid code to any python IDE. It will throw an exception during runtime:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/loadable.py", line 247, in __init__
    raise ValueError("One of num or span must be specified, but not both")
ValueError: One of num or span must be specified, but not both
```

But that can be annoying to deal with. Since `span` is the more powerful of the two options, we should make that a required argument, and probably remove `num` entirely

tybug:
closing as implemented with linked core 4.0.0 release

PR #138, opened 2020-04-19
Core 4.0.0

tybug:
This should have been split into a few prs but it was much easier to do it all at once.

Was thinking about ways we could change circlecore to make it easier to use, and the first thing that came to mind was `Check` - when I use core it's a quick check involving one or two loadables. Obviously I'm not the entire target userbase, but the only thing `Check` does is aggregate all the relevant information together. When are you ever going to need to manipulate a `Check` directly, other than passing it straight to `cg.run`? It's actually more organized if we remove the middleman entirely, in my opinion.

That said, inside core, `Check` *does* provide us some nice convenience methods, like loading or accessing `replays1` or `replays2` instead of the entire check. So using it internally, but not requiring it for `cg.run`, was the path I chose here.

The second thing was creation of `ReplayContainers`. I add "good enough" implementations for `#load` and `#all_loadables`, which makes adding a new `ReplayContainer` feel much less boiler-platey. Here's an example (using this pr):

```python
class RandomReplays(ReplayContainer):
    VALID_MAP_IDS = [110, 145, 234, 420, 734, 1360] # ... etc
    def __init__(self, num_replays, cache=None):
        super().__init__(cache)
        self.map_ids = random.sample(RandomReplays.VALID_MAP_IDS, num_replays)
        self.replays = []

    def load_info(self, loader):
        if self.info_loaded:
            return
        for map_id in self.map_ids:
            # choose a random replay on the map
            infos = loader.replay_info(map_id, Loader.MAX_MAP_SPAN)
            info = random.choice(infos)
            r = ReplayMap(info.map_id, info.user_id, info.mods, cache=self.cache, info=info)
            self.replays.append(r)
        self.info_loaded = True

    def all_replays(self):
        return self.replays

    def __eq__(self, other):
        # don't ever rely on two random replay objects being equal
        return False
```

Something to investigate is whether we even need or want to require `__eq__`. I don't think we use it right now, but if we decided to in the future it would be a breaking change, so some use case consideration is needed.

Third, `span` has now replaced `num` everywhere. It's a little more unwieldy to type - `Map(221777, 2)` vs `Map(221777, "1-2")` - but I feel the tradeoff is worth it for the IDE support users receive when we drop one of the two options. See #115 (which this will close).

The span utils have been wrapped in a new `Span` object, which also allows users to be more literal about creation of `Spans` instead of passing strings to loadables, should they so choose. Additionally, this lets us define spans like `MAX_MAP_SPAN` in `Loader` as a utility for users.

Finally, deprecated objects have been removed as marked.

I'd also like to consider how we can restructure `Detect` to be easier to use. Single use instantiation is common for me, where I create the detect in the same line as I call `cg.run`. Perhaps we should offer `cg.run_steal` etc, since the `Detect`s are nothing more than an encapsulation of thresholds? I'm unsure of the best way forward with this. #132 should also be considered when discussing new `Detect` changes.


tybug:
I'm having quite a bit of trouble writing this new method right now, but I do think we want to require `__eq__`. In the future I'd like to be able to reduce duplicate loading by "minimizing" a `Check`, effectively removing duplicate loadables. This would have to happen through comparing the loadables via `==`.

I added the convenience detect methods as well. Similar to `Check`, I don't want to remove `Detect`, as it's useful both internally in core and for users with more complicated code, using multiple detect combinations at different thresholds. But I like it a lot for the simple use case. It also has the benefit of returning only one type of detect, so 3 sets of `cg.*_check` method calls is (imo) nicer than using `cg.run` and checking the result type

tybug:
Latest set of commits removes any concept of `ischeat` from core, which allows us to turn the previous `Detect` classes into just `IntFlag` enums.

`Detect`s still have data they need to carry with them sometimes, like the `Snap` angle and distance thresholds. It's unfortunate we can't encapsulate these any better without repeating the `Detect` classes. I've left them as parameters to `cg.run` and `cg.correction_check`. Future detects that we want to provide similar value adjustment ability for will be dealt with in a similar way.

As for `ischeat`, the only advantage to having it is to provide users with a sensible default. Which is a good thing to do, and I wish we could replicate it here. The only way I see to do this is provide a threshold parameter in each of the `cg.*_check` methods, and again in `cg.run`. This wouldn't be unreasonable, but I need to consider which is more likely - users adjusting the default thresholds (a pain to do through our core - they should just check the result against their threshold) or not.

circleguard/span.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    def __init__(self, data):
+        # allow passing both span or string
+        if not (isinstance(data, Span) or isinstance(data, str)):
+            raise ValueError(f"Expected data to be a str or Span, got type {type(data)}.")
+        if isinstance(data, Span):
+            # python allows initializing a set with a set
+            super().__init__(data)
+            return
```
InvisibleSymbol, line 19: you don't need this `return`

source/investigating-replays.rst
```diff
@@ -0,0 +1,84 @@
...
+``Circleguard`` for ``App Name`` and
+``https://github.com/circleguard/circlecore`` for ``App URL``. Circlecore
+needs this key to retrieve replay data and user information, among other
+things.
+
+.. note::
+
+    You may need to toy with being logged in/entering the url
+    multiple times in order to actually get to the api page.
+    The redirection on osu!'s end is rather fiddly.
```
InvisibleSymbol, line 20:
```suggestion
    Due to a redirection bug on the website, you might have
    to wait a bit before being able to access the api page. 
    Clearing your cache can be used as a fast alternative, 
    though normally it should only take up to 30 seconds.
```

source/appendix.rst
```diff
@@ -38,24 +38,30 @@ Investigator
...
 Result
 ------
 .. automodule:: circleguard.result
    :members:
 
-ReplayInfo
-----------
-.. automodule:: circleguard.replay_info
+Span
+-----
```
InvisibleSymbol, line 63:
```suggestion
----
```
:eyes:

circleguard/loader.py
```diff
@@ -129,8 +129,16 @@ class Loader():
...
     """
 
-    RATELIMIT_RESET = 60 # time in seconds until the api refreshes our ratelimits
-    start_time = datetime.min # when we started our requests cycle
+    # the maximum number of replay info available through the respective api
+    # calls. Note that osu! stores at least the top 1000 replays, but does not
+    # make these discoverable unless you know the eaact user id, map id, and
+    # mods of the replay.
+    MAX_MAP_SPAN = Span("1-100")
+    MAX_USER_NUM = Span("1-100")
```
InvisibleSymbol, line 137:
```suggestion
    MAX_USER_SPAN = Span("1-100")
```

tybug:
yeah the tests pass and the docs build fine, it's possible I broke something we don't have tests for but I manually made sure all the important things work

PR #139, opened 2020-05-02
Prototyped a correlation metric for replay similarity

kszlim:
Added a correlation metric based on https://en.wikipedia.org/wiki/Cross-correlation#Normalization to detect replay stealing.

Pros over existing metric:
Robust to spatial translation
Robust to outliers (imagine a long replay that is copied and the cheater moves his cursor extensively in the break)
The correlation metric is more widely used in science

Cons:
Different from what already exists
I'm bad at math, so this could be an abuse of signal processing techniques


tybug:
Thanks for the PR! @samuelhklumpers can you review this math-wise? It's over my head I'm afraid.

tests/test.py
```diff
@@ -25,7 +26,7 @@
 class CGTestCase(TestCase):
     @classmethod
     def setUpClass(cls):
-        cls.cg = Circleguard(KEY)
+        cls.cg = Circleguard(KEY, db_path='cache.db')
```
kszlim, line 29: I'll remove this, but is there a reason unit tests can't cache by default? Kinda annoying getting rate limited if you run the unit tests a couple times in a row (and this I think not everything is getting cached, so I still get rate limited even with this enabled).

kszlim:
Also probably would be good to have a large suite of known copied/slightly modified replays so we can get definitive metrics about how good these replay stealing analysis techniques are.

tests/test.py
tybug: no reason at all - we should cache and just gitignore the db. Having that should stop any ratelimiting from happening though. I think the reason it's not is because we create a new `Circleguard` instance in most of the test cases. 

I think this is a holdover from when I was more worried about `Circleguard` saving state in some way and wanted to start fresh each time. That should no longer be a concern and the repeated instantiation can be removed

circleguard/loadable.py
```diff
@@ -722,3 +722,8 @@ def __str__(self):
             return f"Loaded ReplayPath by {self.username} on {self.map_id} at {self.path}"
         else:
             return f"Unloaded ReplayPath at {self.path}"
+
+    def positionally_translate_replay(self):
```
kszlim, line 726: Would've preferred to put this in the test file, but unfortunately, loading of replay data is lazy, so it only loads once you start an analysis. It also resulted in me putting in a bit of a hack to enable positional translation + noise when a certain flag is marked.

tybug: you can force replay data loading with `cg.load(replay)`, in case you didn't know

circleguard/comparer.py
```diff
@@ -44,6 +45,9 @@ def __init__(self, replays1, replays2=None):
 
         # filter beatmaps we had no data for
         self.replays1 = [replay for replay in replays1 if replay.replay_data is not None]
+        for replay in self.replays1:
+            if hasattr(replay, 'should_translate'):
```
kszlim, line 49: This is the hack I put in to enable the test. Not sure if there's a better way due to the lazy loading.

tybug:
I actually have just that (37 pairs of stolen replays), but haven't put them in the tests yet. Been meaning to for a while. I think this is the PR that makes me clean up our tests a bit.

circleguard/loadable.py
kszlim: Sounds good, where do you prefer `positionally_translate_replay` to live?

kszlim: and also, would you want it parameterized? A lot of what i through together was kinda a hack.

circleguard/loadable.py
```diff
@@ -722,3 +722,8 @@ def __str__(self):
...
+        noise = np.random.normal(0, 1, )
+        self.xy = self.xy + np.random.default_rng().multivariate_normal([10, 10], [[3,3],[3,3]],(len(self.xy)))
```
tybug, line 728: What version of numpy are you running? This errors for me -

```python
Traceback (most recent call last):
  File "/Users/tybug/Desktop/coding/osu/circlecore/tests/test.py", line 72, in test_robustness_to_translation
    r = list(self.cg.steal_check(replays))
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/circleguard.py", line 151, in steal_check
    yield from self.run(loadables, Detect.STEAL, loadables2)
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/circleguard.py", line 107, in run
    comparer = Comparer(replays1, replays2=replays2)
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py", line 50, in __init__
    replay.positionally_translate_replay()
  File "/Users/tybug/Desktop/coding/osu/circlecore/circleguard/loadable.py", line 728, in positionally_translate_replay
    self.xy = self.xy + np.random.default_rng().multivariate_normal([10, 10], [[3,3],[3,3]],(len(self.xy)))
AttributeError: module 'numpy.random' has no attribute 'default_rng'
```

I'm on `1.15.4`. I don't think we pin numpy to a minimum version in `setup.py`, but maybe we should.

kszlim: This is my pipfreeze:

```
circleparse==6.2.0
click==7.1.2
idna==2.9
numpy==1.18.3
ossapi==1.2.3
pytz==2020.1
requests==2.23.0
scipy==1.4.1
slider==0.2.1
toolz==0.10.0
urllib3==1.25.9
wtc==1.1.3
```

circleguard/loadable.py
tybug: This is just for testing right? Should probably live in the tests file then, like you suggested

kszlim:
With respect to:

```
I actually have just that (37 pairs of stolen replays), but haven't put them in the tests yet. Been meaning to for a while. I think this is the PR that makes me clean up our tests a bit.
```

It might be even better if we can get a large corpus of cheated replays from @peppy which might allow application of ML techniques or even for optimization of parameters currently in circlecore.

circleguard/comparer.py
```diff
@@ -144,7 +149,33 @@ def _compare_two_replays(replay1, replay2):
         if (Mod.HR in replay1.mods) ^ (Mod.HR in replay2.mods):
             xy1[:, 1] = 384 - xy1[:, 1]
 
-        return Comparer._compute_data_similarity(xy1, xy2)
+        (play_sequence_1, play_sequence_2) = Comparer.convert_to_row_major_form(xy1, xy2)
+
+        # Sectioned into 20 chunks, used to ignore outliers (eg. long replay with breaks is copied, and the cheater cursordances during the break)
```
kszlim, line 154: 20 was arbitrarily chosen, whether to section maps into constant chunks or a constant number of chunks is also up for debate.

kszlim:
Read through my numpy code carefully,  I don't do much scientific computing, so there probably are a lot of bad practices.

circleguard/loadable.py
tybug: Looks like an addition in `1.17` or later: <https://github.com/numpy/numpy/pull/13873>, unless I'm misinterpreting this pr.

tybug:
> It might be even better if we can get a large corpus of cheated replays from @ peppy which might allow application of ML techniques or even for optimization of parameters currently in circlecore.

I doubt peppy would be willing to release any information on restricted users that isn't already publicly known, given the edge it could give cheat developers if they knew what replays were being detected.

circleguard/result.py
```diff
@@ -74,12 +74,16 @@ class StealResult(ComparisonResult):
         The similarity of the two replays (the lower, the more similar).
         Similarity is, roughly speaking, a measure of the average pixel
         distance between the two replays.
+    correlation: float
+        The median correlation of the two replays (the closer to 1, the more correlated).
+        A value of 1.0 would be perfectly correlated, and a result of 0 would be uncorrelated.
     """
 
-    def __init__(self, replay1: Replay, replay2: Replay, similarity: int,):
+    def __init__(self, replay1: Replay, replay2: Replay, similarity: int, correlation: float=0.0):
```
kszlim, line 82: default param for backwards compatibility

tests/test.py
kszlim: any preferred path for the cache then?

circleguard/comparer.py
kszlim: Even for tiny maps, I think we have enough data points such that slicing it into chunks of size 20 is likely still okay.

Sample rate is generally about 40-60hz.

tests/test.py
tybug: the tests directory - `cache_path = Path(__file__).parent`

circleguard/comparer.py
```diff
@@ -118,8 +119,9 @@ def _compare_two_replays(replay1, replay2):
 
         Returns
         -------
-        float
-            The mean distance of the cursors of the two replays.
+        (float, float)
+            The mean distance of the cursors of Use the API for good. Don't overdo it. If in doubt, ask before (ab)using :). this section may expand as necessary.
```
samuelhklumpers, line 123: what is this `Use the API for good. Don't overdo it. If in doubt, ask before (ab)using :). this section may expand as necessary.`

circleguard/comparer.py
```diff
@@ -144,7 +146,35 @@ def _compare_two_replays(replay1, replay2):
...
+            #Pick the lag with the maximum correlation, this likely in most cases is 0 lag
+            max_correlation = max(cross_correlation_matrix.reshape(cross_correlation_matrix.size))
+            correlations.append(max_correlation)
+        average_distance = Comparer._compute_data_similarity(xy1, xy2)
+        # Out of all the sections, we should have 20 sections, we pick the one with the median correlation, so we throw away any outliers
+        median_correlation = np.median(np.array(sorted(correlations)))
+        return (average_distance, median_correlation)
+
+    @staticmethod
+    def convert_to_row_major_form(x, y):
```
samuelhklumpers, line 169: this could be replaced with `x.T` in all usages

circleguard/comparer.py
```diff
@@ -144,7 +146,35 @@ def _compare_two_replays(replay1, replay2):
...
             xy1[:, 1] = 384 - xy1[:, 1]
 
-        return Comparer._compute_data_similarity(xy1, xy2)
+        (play_sequence_1, play_sequence_2) = Comparer.convert_to_row_major_form(xy1, xy2)
+
+        NUM_CHUNKS = 20
+
+        # Sectioned into 20 chunks, used to ignore outliers (eg. long replay with breaks is copied, and the cheater cursordances during the break)
+        horizontal_length = play_sequence_1.shape[1] - play_sequence_1.shape[1] % NUM_CHUNKS
+        play_sequence_1_sections = np.hsplit(play_sequence_1[:,:horizontal_length], NUM_CHUNKS)
```
samuelhklumpers, line 155: after interpolation both replays are equally long so you might be able to include the remainder section in the cross correlation

circleguard/comparer.py
```diff
@@ -144,7 +146,35 @@ def _compare_two_replays(replay1, replay2):
...
+        return (average_distance, median_correlation)
+
+    @staticmethod
+    def convert_to_row_major_form(x, y):
+        return (np.array(list(zip(*x))), np.array(list(zip(*y))))
+
+    @staticmethod
+    def find_cross_correlation(x, y):
+        # The normalization is makes the similarity detector robust to translations
+        x = (x - np.mean(x)) / (np.std(x) * x.size)
```
samuelhklumpers, line 175: you should probably do this normalization after the cross correlation, as this normalizes all data when doing it after only normalizes 1 number

circleguard/comparer.py
```diff
@@ -144,7 +146,35 @@ def _compare_two_replays(replay1, replay2):
...
+
+        # Sectioned into 20 chunks, used to ignore outliers (eg. long replay with breaks is copied, and the cheater cursordances during the break)
+        horizontal_length = play_sequence_1.shape[1] - play_sequence_1.shape[1] % NUM_CHUNKS
+        play_sequence_1_sections = np.hsplit(play_sequence_1[:,:horizontal_length], NUM_CHUNKS)
+        play_sequence_2_sections = np.hsplit(play_sequence_2[:,:horizontal_length], NUM_CHUNKS)
+        correlations = []
+        for (play_sequence_1, play_sequence_2) in zip(play_sequence_1_sections, play_sequence_2_sections):
+            cross_correlation_matrix = Comparer.find_cross_correlation(play_sequence_1, play_sequence_2)
+            #Pick the lag with the maximum correlation, this likely in most cases is 0 lag
+            max_correlation = max(cross_correlation_matrix.reshape(cross_correlation_matrix.size))
```
samuelhklumpers, line 161: is this reshape necessary if you use numpy's max?

samuelhklumpers: max might be good to use but unless there's a good reason against it, consider using a mean, as max has an upward bias and might give some unfortunate outliers

circleguard/comparer.py
kszlim: oh lol accidentally copy pasta'd that into the file from Peppy's documentation on the api.

circleguard/comparer.py
kszlim: Yeah, I could, you think it's worth it for the slightly increased code complexity?

kszlim: Also, got any feedback around fix chunk sizes vs fixed number of chunks? Based on the sample rate, I think it should be okay for any ranked maps, but who knows.

circleguard/comparer.py
kszlim: Hmm, maybe my understanding of the math is wrong, but this what I'm basing it off.

https://www.statisticshowto.com/cross-correlation/

The max of the cross correlation is where the two signals correlate the best. The mean of the cross correlation matrix will almost always be a very low number afaik.

Imagine if you get the correlations of sin(t) and sin(t - n), so you end up with two signals that are the same but out of phase. You'll get a normalized correlation matrix where the mean will be 0 despite the signals being highly correlated. Picking the max from the cross correlation matrix will give you a correlation of 1, telling you that the signals are the same, but shifted in the time dimension.

I could be wrong though, honestly, haven't done any math in years.

I think the outlier problem is actually solved by picking the median out of all the max correlations from every single slice?

samuelhklumpers: you are correct and you should use max (it doesn't introduce bias because it's the only useful value to inspect), I had mistakenly imagined the correlation as being instantaneous but it is not and that's the only reason you need to even divide the replay. 
The way you do correlation should also filter shifts in the xy-plane correct?

circleguard/comparer.py
samuelhklumpers: I don't think the code would get very much more complex. I also have no strong opinion on fixed/variable chunks but I am not opposed to it being a default variable with value 20

circleguard/comparer.py
kszlim: Yes, it should catch translations in the xy plane.

circleguard/comparer.py
kszlim: Hmm, might not be worth adding, as you could potentially end up with a slice with just 1 data point, for which this would fail on.

Will parameterize the number of chunks though and default it to 20.

circleguard/comparer.py
kszlim: Hmm, still unsure what you mean, maybe post some code, or make the changes you want to this pr?

samuelhklumpers: something like
```
x -= np.mean(x)
y -= np.mean(y)
norm = np.std(x) * np.std(y) * x.size
return signal.correlate(x, y) / norm
```

circleguard/comparer.py
```diff
@@ -144,7 +146,32 @@ def _compare_two_replays(replay1, replay2):
         if (Mod.HR in replay1.mods) ^ (Mod.HR in replay2.mods):
             xy1[:, 1] = 384 - xy1[:, 1]
 
-        return Comparer._compute_data_similarity(xy1, xy2)
+        play_sequence_1 = xy1.T.copy() # Need copy, as the cross_correlate method uses an in-place modification of the matrix
+        play_sequence_2 = xy2.T.copy()
```
tybug, line 150: any reason you can't just use `xy1` and `xy2` here?

circleguard/comparer.py
```diff
@@ -144,7 +146,32 @@ def _compare_two_replays(replay1, replay2):
...
+
+        # Sectioned into 20 chunks, used to ignore outliers (eg. long replay with breaks is copied, and the cheater cursordances during the break)
+        horizontal_length = play_sequence_1.shape[1] - play_sequence_1.shape[1] % num_chunks
+        play_sequence_1_sections = np.hsplit(play_sequence_1[:,:horizontal_length], num_chunks)
+        play_sequence_2_sections = np.hsplit(play_sequence_2[:,:horizontal_length], num_chunks)
+        correlations = []
+        for (play_sequence_1, play_sequence_2) in zip(play_sequence_1_sections, play_sequence_2_sections):
+            cross_correlation_matrix = Comparer.find_cross_correlation(play_sequence_1, play_sequence_2)
+            #Pick the lag with the maximum correlation, this likely in most cases is 0 lag
+            max_correlation = max(cross_correlation_matrix.reshape(cross_correlation_matrix.size))
```
tybug, line 160: using `np.max` should avoid the reshape here

circleguard/comparer.py
```diff
@@ -144,7 +146,32 @@ def _compare_two_replays(replay1, replay2):
...
+            correlations.append(max_correlation)
+        average_distance = Comparer._compute_data_similarity(xy1, xy2)
+        # Out of all the sections, we should have 20 sections, we pick the one with the median correlation, so we throw away any outliers
+        median_correlation = np.median(np.array(sorted(correlations)))
+        return (average_distance, median_correlation)
+
+    @staticmethod
+    def find_cross_correlation(play_sequence_1, play_sequence_2):
```
tybug, line 168: rename to `xy1` and `xy2`

`_compute_data_similarity` params should also be renamed to this

circleguard/comparer.py
tybug: this was a vague comment, sorry - what I really meant was it should be named something with `xy1` and `xy2` in the name, not `play_sequence`

tybug:
Unfortunately, this takes quite a bit to run. Here's running both similarity and correlation on two replays 100 times (5.42 seconds):

```
Timer unit: 1e-06 s

Total time: 5.42588 s
File: /Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py
Function: _compare_two_replays at line 107

   107                                               @staticmethod
   108                                               @profile
   109                                               def _compare_two_replays(replay1, replay2, num_chunks=20):
   110                                                   """
   111                                                   Calculates the average cursor distance between two
   112                                                   :class:`~.replay.Replay`\s, and the standard deviation of the distance.
   113                                           
   114                                                   Parameters
   115                                                   ----------
   116                                                   replay1: :class:`~.replay.Replay`
   117                                                       The first replay to compare.
   118                                                   replay2: :class:`~.replay.Replay`
   119                                                       The second replay to compare.
   120                                           
   121                                                   Returns
   122                                                   -------
   123                                                   (float, float)
   124                                                       The mean distance of the cursors of the two replays.
   125                                                       Also returns a correlation metric that represents the median level of correlation in the replay.
   126                                                   """
   127                                           
   128                                                   # interpolate
   129       100        283.0      2.8      0.0          if len(replay1.t) > len(replay2.t):
   130                                                       xy1 = np.array(replay1.xy)  # implicit fast copy to avoid overwriting this array on replay1 when flipping for HR
   131                                           
   132                                                       xy2x = np.interp(replay1.t, replay2.t, replay2.xy[:, 0])
   133                                                       xy2y = np.interp(replay1.t, replay2.t, replay2.xy[:, 1])
   134                                                       xy2 = np.array([xy2x, xy2y]).T
   135                                                   else:
   136       100      64301.0    643.0      1.2              xy1x = np.interp(replay2.t, replay1.t, replay1.xy[:, 0])
   137       100      45237.0    452.4      0.8              xy1y = np.interp(replay2.t, replay1.t, replay1.xy[:, 1])
   138       100       4488.0     44.9      0.1              xy1 = np.array([xy1x, xy1y]).T
   139                                           
   140       100       5760.0     57.6      0.1              xy2 = np.array(replay2.xy)  # implicit fast copy to avoid overwriting this array on replay2 when flipping for HR
   141                                           
   142       100      85628.0    856.3      1.6          valid = np.all(([0, 0] <= xy1) & (xy1 <= [512, 384]), axis=1) & np.all(([0, 0] <= xy2) & (xy2 <= [512, 384]), axis=1)
   143       100      46772.0    467.7      0.9          xy1 = xy1[valid]
   144       100      45949.0    459.5      0.8          xy2 = xy2[valid]
   145                                           
   146                                                   # flip if one but not both has HR
   147       100       1141.0     11.4      0.0          if (Mod.HR in replay1.mods) ^ (Mod.HR in replay2.mods):
   148                                                       xy1[:, 1] = 384 - xy1[:, 1]
   149                                           
   150       100       7358.0     73.6      0.1          # play_sequence_1 = xy1.T.copy() # Need copy, as the cross_correlate method uses an in-place modification of the matrix
   151       100       7006.0     70.1      0.1          # play_sequence_2 = xy2.T.copy()
   152                                           
   153                                                   # # Sectioned into 20 chunks, used to ignore outliers (eg. long replay with breaks is copied, and the cheater cursordances during the break)
   154       100        308.0      3.1      0.0          # horizontal_length = play_sequence_1.shape[1] - play_sequence_1.shape[1] % num_chunks
   155       100      20050.0    200.5      0.4          # play_sequence_1_sections = np.hsplit(play_sequence_1[:,:horizontal_length], num_chunks)
   156       100      14427.0    144.3      0.3          # play_sequence_2_sections = np.hsplit(play_sequence_2[:,:horizontal_length], num_chunks)
   157       100        100.0      1.0      0.0          # correlations = []
   158      2100       3459.0      1.6      0.1          # for (play_sequence_1, play_sequence_2) in zip(play_sequence_1_sections, play_sequence_2_sections):
   159      2000    3130086.0   1565.0     57.7          #     cross_correlation_matrix = Comparer.find_cross_correlation(play_sequence_1, play_sequence_2)
   160                                                   #     #Pick the lag with the maximum correlation, this likely in most cases is 0 lag
   161      2000    1848423.0    924.2     34.1          #     max_correlation = max(cross_correlation_matrix.reshape(cross_correlation_matrix.size))
   162      2000       4753.0      2.4      0.1          #     correlations.append(max_correlation)
   163       100      71979.0    719.8      1.3          average_distance = Comparer._compute_data_similarity(xy1, xy2)
   164                                                   # # Out of all the sections, we should have 20 sections, we pick the one with the median correlation, so we throw away any outliers
   165       100      18256.0    182.6      0.3          # median_correlation = np.median(np.array(sorted(correlations)))
   166       100        112.0      1.1      0.0          # return (average_distance, median_correlation)
   167                                                   return average_distance

Total time: 3.09533 s
File: /Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py
Function: find_cross_correlation at line 168

   168                                           
   169                                               # @staticmethod
   170                                               # @profile
   171                                               # def find_cross_correlation(play_sequence_1, play_sequence_2):
   172      2000     151475.0     75.7      4.9      #     # The normalization makes the similarity detector robust to translations
   173      2000      97105.0     48.6      3.1      #     play_sequence_1 -= np.mean(play_sequence_1)
   174      2000     292511.0    146.3      9.5      #     play_sequence_2 -= np.mean(play_sequence_2)
   175      2000    2554243.0   1277.1     82.5      #     norm = np.std(play_sequence_1) * np.std(play_sequence_2) * play_sequence_1.size
   176                                               #     return signal.correlate(play_sequence_1, play_sequence_2) / norm
   177                                           
   178                                           
   179                                               @staticmethod
   180                                               @profile
   181                                               def _compute_data_similarity(data1, data2):
   182                                                   """
   183                                                   Calculates the average cursor distance between two lists of cursor data,
   184                                                   and the standard deviation of the distance.
   185                                           
   186                                                   Parameters
   187                                                   ----------
   188                                                   data1: list[tuple(int, int)]
   189                                                       The first set of cursor data, containing the x and y positions
   190                                                       of the cursor at each datapoint.
   191                                                   data2: list[tuple(int, int)]
   192                                                       The second set of cursor data, containing the x and y positions
   193                                                       of the cursor at each datapoint.
   194                                           
   195                                                   Returns
   196                                                   -------
   197                                                   float
   198                                                       The average distance between the two datasets.
   199                                           
   200                                                   Notes
   201                                                   -----
   202                                                   The two data lists must have previously been interpolated to each other.
   203                                                   This is why we can get away with only lists of [x,y] and not [x,y,t].
   204                                                   """
   205                                           
   206                                                   data1 = np.array(data1)
   207                                                   data2 = np.array(data2)
   208                                           
   209                                                   # switch if the second is longer, so that data1 is always the longest.
   210                                                   if len(data2) > len(data1):
   211                                                       (data1, data2) = (data2, data1)
   212                                           
   213                                                   shortest = len(data2)
   214                                           
   215                                                   distance = data1[:shortest] - data2
   216                                                   # square all numbers and sum over the second axis (add row 2 to row 1),
   217                                                   # finally take the square root of each number to get all distances.
   218                                                   # [ x_1 x_2 ... x_n   => [ x_1 ** 2 ... x_n ** 2
   219                                                   #   y_1 y_2 ... y_n ] =>   y_1 ** 2 ... y_n ** 2 ]
   220                                                   # => [ x_1 ** 2 + y_1 ** 2 ... x_n ** 2 + y_n ** 2 ]
   221                                                   # => [ d_1 ... d_2 ]
   222                                                   distance = (distance ** 2).sum(axis=1) ** 0.5
   223                                           
   224                                                   return distance.mean()

Total time: 0.069345 s
File: /Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py
Function: _compute_data_similarity at line 178

   178                                           
   179                                               @staticmethod
   180                                               @profile
   181                                               def _compute_data_similarity(data1, data2):
   182                                                   """
   183                                                   Calculates the average cursor distance between two lists of cursor data,
   184                                                   and the standard deviation of the distance.
   185                                           
   186                                                   Parameters
   187                                                   ----------
   188                                                   data1: list[tuple(int, int)]
   189                                                       The first set of cursor data, containing the x and y positions
   190                                                       of the cursor at each datapoint.
   191                                                   data2: list[tuple(int, int)]
   192                                                       The second set of cursor data, containing the x and y positions
   193                                                       of the cursor at each datapoint.
   194                                           
   195                                                   Returns
   196                                                   -------
   197                                                   float
   198                                                       The average distance between the two datasets.
   199                                           
   200                                                   Notes
   201                                                   -----
   202                                                   The two data lists must have previously been interpolated to each other.
   203                                                   This is why we can get away with only lists of [x,y] and not [x,y,t].
   204                                                   """
   205       100       5880.0     58.8      8.5  
   206       100       5455.0     54.5      7.9          data1 = np.array(data1)
   207                                                   data2 = np.array(data2)
   208                                           
   209       100        203.0      2.0      0.3          # switch if the second is longer, so that data1 is always the longest.
   210                                                   if len(data2) > len(data1):
   211                                                       (data1, data2) = (data2, data1)
   212       100         58.0      0.6      0.1  
   213                                                   shortest = len(data2)
   214       100       6513.0     65.1      9.4  
   215                                                   distance = data1[:shortest] - data2
   216                                                   # square all numbers and sum over the second axis (add row 2 to row 1),
   217                                                   # finally take the square root of each number to get all distances.
   218                                                   # [ x_1 x_2 ... x_n   => [ x_1 ** 2 ... x_n ** 2
   219                                                   #   y_1 y_2 ... y_n ] =>   y_1 ** 2 ... y_n ** 2 ]
   220                                                   # => [ x_1 ** 2 + y_1 ** 2 ... x_n ** 2 + y_n ** 2 ]
   221       100      46495.0    464.9     67.0          # => [ d_1 ... d_2 ]
   222                                                   distance = (distance ** 2).sum(axis=1) ** 0.5
   223       100       4741.0     47.4      6.8  
   224                                                   return distance.mean()
```

And here's running just similarity on the same two replays 100 times (0.3 seconds):

```
Timer unit: 1e-06 s

Total time: 0.309009 s
File: /Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py
Function: _compare_two_replays at line 107

   107                                               @staticmethod
   108                                               @profile
   109                                               def _compare_two_replays(replay1, replay2, num_chunks=20):
   110                                                   """
   111                                                   Calculates the average cursor distance between two
   112                                                   :class:`~.replay.Replay`\s, and the standard deviation of the distance.
   113                                           
   114                                                   Parameters
   115                                                   ----------
   116                                                   replay1: :class:`~.replay.Replay`
   117                                                       The first replay to compare.
   118                                                   replay2: :class:`~.replay.Replay`
   119                                                       The second replay to compare.
   120                                           
   121                                                   Returns
   122                                                   -------
   123                                                   (float, float)
   124                                                       The mean distance of the cursors of the two replays.
   125                                                       Also returns a correlation metric that represents the median level of correlation in the replay.
   126                                                   """
   127                                           
   128                                                   # interpolate
   129       100        168.0      1.7      0.1          if len(replay1.t) > len(replay2.t):
   130                                                       xy1 = np.array(replay1.xy)  # implicit fast copy to avoid overwriting this array on replay1 when flipping for HR
   131                                           
   132                                                       xy2x = np.interp(replay1.t, replay2.t, replay2.xy[:, 0])
   133                                                       xy2y = np.interp(replay1.t, replay2.t, replay2.xy[:, 1])
   134                                                       xy2 = np.array([xy2x, xy2y]).T
   135                                                   else:
   136       100      41322.0    413.2     13.4              xy1x = np.interp(replay2.t, replay1.t, replay1.xy[:, 0])
   137       100      40761.0    407.6     13.2              xy1y = np.interp(replay2.t, replay1.t, replay1.xy[:, 1])
   138       100       3951.0     39.5      1.3              xy1 = np.array([xy1x, xy1y]).T
   139                                           
   140       100       3648.0     36.5      1.2              xy2 = np.array(replay2.xy)  # implicit fast copy to avoid overwriting this array on replay2 when flipping for HR
   141                                           
   142       100      77793.0    777.9     25.2          valid = np.all(([0, 0] <= xy1) & (xy1 <= [512, 384]), axis=1) & np.all(([0, 0] <= xy2) & (xy2 <= [512, 384]), axis=1)
   143       100      40846.0    408.5     13.2          xy1 = xy1[valid]
   144       100      42592.0    425.9     13.8          xy2 = xy2[valid]
   145                                           
   146                                                   # flip if one but not both has HR
   147       100        772.0      7.7      0.2          if (Mod.HR in replay1.mods) ^ (Mod.HR in replay2.mods):
   148                                                       xy1[:, 1] = 384 - xy1[:, 1]
   149                                           
   150                                                   # play_sequence_1 = xy1.T.copy() # Need copy, as the cross_correlate method uses an in-place modification of the matrix
   151                                                   # play_sequence_2 = xy2.T.copy()
   152                                           
   153                                                   # # Sectioned into 20 chunks, used to ignore outliers (eg. long replay with breaks is copied, and the cheater cursordances during the break)
   154                                                   # horizontal_length = play_sequence_1.shape[1] - play_sequence_1.shape[1] % num_chunks
   155                                                   # play_sequence_1_sections = np.hsplit(play_sequence_1[:,:horizontal_length], num_chunks)
   156                                                   # play_sequence_2_sections = np.hsplit(play_sequence_2[:,:horizontal_length], num_chunks)
   157                                                   # correlations = []
   158                                                   # for (play_sequence_1, play_sequence_2) in zip(play_sequence_1_sections, play_sequence_2_sections):
   159                                                   #     cross_correlation_matrix = Comparer.find_cross_correlation(play_sequence_1, play_sequence_2)
   160                                                   #     #Pick the lag with the maximum correlation, this likely in most cases is 0 lag
   161                                                   #     max_correlation = max(cross_correlation_matrix.reshape(cross_correlation_matrix.size))
   162                                                   #     correlations.append(max_correlation)
   163       100      57096.0    571.0     18.5          average_distance = Comparer._compute_data_similarity(xy1, xy2)
   164                                                   # # Out of all the sections, we should have 20 sections, we pick the one with the median correlation, so we throw away any outliers
   165                                                   # median_correlation = np.median(np.array(sorted(correlations)))
   166                                                   # return (average_distance, median_correlation)
   167       100         60.0      0.6      0.0          return average_distance

Total time: 0.055162 s
File: /Users/tybug/Desktop/coding/osu/circlecore/circleguard/comparer.py
Function: _compute_data_similarity at line 179

   179                                               @staticmethod
   180                                               @profile
   181                                               def _compute_data_similarity(data1, data2):
   182                                                   """
   183                                                   Calculates the average cursor distance between two lists of cursor data,
   184                                                   and the standard deviation of the distance.
   185                                           
   186                                                   Parameters
   187                                                   ----------
   188                                                   data1: list[tuple(int, int)]
   189                                                       The first set of cursor data, containing the x and y positions
   190                                                       of the cursor at each datapoint.
   191                                                   data2: list[tuple(int, int)]
   192                                                       The second set of cursor data, containing the x and y positions
   193                                                       of the cursor at each datapoint.
   194                                           
   195                                                   Returns
   196                                                   -------
   197                                                   float
   198                                                       The average distance between the two datasets.
   199                                           
   200                                                   Notes
   201                                                   -----
   202                                                   The two data lists must have previously been interpolated to each other.
   203                                                   This is why we can get away with only lists of [x,y] and not [x,y,t].
   204                                                   """
   205                                           
   206       100       3282.0     32.8      5.9          data1 = np.array(data1)
   207       100       4222.0     42.2      7.7          data2 = np.array(data2)
   208                                           
   209                                                   # switch if the second is longer, so that data1 is always the longest.
   210       100        173.0      1.7      0.3          if len(data2) > len(data1):
   211                                                       (data1, data2) = (data2, data1)
   212                                           
   213       100         52.0      0.5      0.1          shortest = len(data2)
   214                                           
   215       100       4716.0     47.2      8.5          distance = data1[:shortest] - data2
   216                                                   # square all numbers and sum over the second axis (add row 2 to row 1),
   217                                                   # finally take the square root of each number to get all distances.
   218                                                   # [ x_1 x_2 ... x_n   => [ x_1 ** 2 ... x_n ** 2
   219                                                   #   y_1 y_2 ... y_n ] =>   y_1 ** 2 ... y_n ** 2 ]
   220                                                   # => [ x_1 ** 2 + y_1 ** 2 ... x_n ** 2 + y_n ** 2 ]
   221                                                   # => [ d_1 ... d_2 ]
   222       100      38619.0    386.2     70.0          distance = (distance ** 2).sum(axis=1) ** 0.5
   223                                           
   224       100       4098.0     41.0      7.4          return distance.mean()
```

Notably the `np.reshape` takes a significant amount of time, something I didn't expect. This should be solved by using `np.max` instead. But `find_cross_correlation` took `3.09533` seconds to run `2000` times. It gets hit 20 times for each comparison because we split the replay into 20 chunks. 

I don't know if the runtime is due to some constant cost involved in invoking the function, or is due to the data size - if the former we can reduce runtime by reducing chunk size, and if the latter we're kind of boned.

tybug:
Chunking isn't necessarily unique to this algorithm, I think - we could do the same with our current method, and it would probably increase runtime by a similar factor. That said, there are alternative approaches, like cutting out xy data in breaks period, which reaches the same goal and is much faster. *That* said, we'd need the beatmap for every replay, or some way to determine the breaks ourselves. Which is why we don't currently do it.

tybug:
I've satisfied my fear that this algo cannot distinguish stolen and non-stolen replays - it does so quite well, with the threshold seemingly at `0.99`.

Random samples off a few maps (~5) from invis' cache: https://pastebin.com/raw/EyMEjEAT
full top 50 on 221777: https://pastebin.com/raw/isMdjgbV

corr goes as high as `0.987` on 221777, but mostly landing solidly in the `0.97` - `0.985` range.

tybug:
Also does equally as well on time shifts as normally stolen replays, though I only tested this with a sample size of n=1.

tybug:
this pr really boils down to

pro - detects time shifts, xy shifts (though I haven't tested this myself), and potentially both combined

con - slow

to me, catching different attack vectors in a different way makes this worth merging, even if it's slower. Gives us options in the future for anything that might come up. I'll get this pr into a mergable state sometime soon, hopefully.

kszlim:
Sounds good, let me know if you need anything from me.

tybug:
this pr will add scipy to our requirements. What's the standard for importing both numpy and scipy? My understanding is that scipy is a superset of numpy, so I don't see why we can't just import things we need from scipy instead and save a line. Do people generally import both in the same file anyway for clarity? cc @samuelhklumpers as well

kszlim:
Not sure, I think I added that when I messed up and didn't source from my venv. Pretty sure it was unnecessary.

tybug:
Seems necessary to me? you use `https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.correlate.html`

kszlim:
Ah yeah, I guess so. Didn't read into it carefully, guess numpy isn't necessary in this case.

tybug:
well we use numpy for other things in the same file so the import needs to stay (in some fashion), my question is whether people move the imports to scipy instead of importing from both numpy and scipy.

kszlim:
Afaik, numpy is more supposed to be a lower level library for matrix operations, and scipy contains algorithms that are at a higher level. Generally if you're doing basic matrix manipulation, you stick to np, and if you're using higher level algorithms, then you use scipy. There's overlap due to historical reasons, but that's the guidance I've read on the internet.

I meant you could probably remove numpy from the setup.py since numpy is probably a transitive dep of scipy and that way you won't accidentally have an incompatible version of numpy and scipy in the same build.

tybug:
ah, my bad. Yeah that's a good point. Will wait and see what sam says.

circleguard/comparer.py
```diff
@@ -100,97 +103,126 @@ def _result(self, replay1, replay2):
...
+        xy1_sections = np.hsplit(xy1[:,:horizontal_length], num_chunks)
+        xy2_sections = np.hsplit(xy2[:,:horizontal_length], num_chunks)
+        correlations = []
+        for (xy1_section, xy2_section) in zip(xy1_sections, xy2_sections):
+            xy1_section -= np.mean(xy1_section)
+            xy2_section -= np.mean(xy2_section)
+            norm = np.std(xy1_section) * np.std(xy2_section) * xy1_section.size
+            cross_correlation_matrix = signal.correlate(xy1_section, xy2_section) / norm
+            # Pick the lag with the maximum correlation, this likely in
+            # most cases is 0 lag
```
tybug, line 167: what does 0 lag mean? what is lag? I'm having a hard time parsing this comment in general.

kszlim: When you take the cross correlation of two identical signals that are out of phase, how much they are out of phase by is the lag of one signal wrt the other signal.

Because we're looking at two signals that are likely very similar, when we take the cross correlation we're given a matrix that measures the normalized cross correlation at various shifts in the time domain. In most cases, our two signals will be unshifted in the time domain relative to each other, and as such, in most cases grabbing the value with the highest correlation from this matrix is also correspondingly picking the the correlation associated with zero shift in the time domain.

So the latter part of the comment is really just to outline that in most cases, the maximum correlation happens we consider the two signals unshifted relative to each other.

tybug: I see, thank you.

How large is the maximal time shift `signal.correlate` checks? Is it possible to bound it to `t ∈ [-2000, 2000]` ms or so, and save processing time? Because we can reasonably discard very large timeshifts as impossible to generate a high scoring replay (ie an FC)

kszlim: Not without essentially implementing a version of signal.correlate. I think you would rewrite that method, but use less zero padding in this step https://github.com/scipy/scipy/blob/v1.4.1/scipy/signal/signaltools.py#L240

Alternatively, I did some testing, and we could use

```python
cross_corr_matrix = signal.correlate(xy1_part, xy2_part, mode='valid', method='direct') / norm
```

which is dramatically faster than currently (about 5-6x). But if I'm interpreting this correlate, I think it kills the ability to detect time shifts.

You can test it with the shifted test case you created to see if that's the case.

tybug: That speeds it up to within 25% or so of the current similarity method for me, but you're right that it doesn't detect time shifts - the one I tested, anyway. Oh well, looks like there's nothing to be gained here then, unless we were to take a deep dive into creating our own correlation for potentially little benefit.

Keeping `method="direct"` in the back of my mind as a potential fast-enough alternative to the current similarity though, if something were to ever go wrong with it.

kszlim: How did you benchmark it? I purely benchmarked the inner loop.

eg.

```python
        for (xy1_part, xy2_part) in zip(xy1_parts, xy2_parts):
            xy1_part -= np.mean(xy1_part)
            xy2_part -= np.mean(xy2_part)
            norm = np.std(xy1_part) * np.std(xy2_part) * xy1_part.size
            cross_corr_matrix = signal.correlate(xy1_part, xy2_part) / norm
            # Pick the lag with the maximum correlation, this likely in
            # most cases is 0 lag
            max_corr = np.max(cross_corr_matrix)
            correlations.append(max_corr)
```

done a thousand times.

Does the current method work against that time shifted replay? I'm actually still kinda unconvinced about the math that's (ab)used to detect time shifts with cross correlation.

tybug: I benchmarked quite a bit outside that, but I'd be surprised if there was a (relative) time difference in moving the benchmarking towards the inside.

```

r1 = ReplayPath("/Users/tybug/Desktop/replays/stealing/stolen/stolen-time-shifted-1.osr")
r2 = ReplayPath("/Users/tybug/Desktop/replays/stealing/stolen/stolen-time-shifted-2.osr")


import time
t0 = time.time()
for _ in range(500):
    list(cg.steal_check([r1, r2], method=Detect.STEAL_CORR))
t1 = time.time()
print(t1 - t0)
``

The current method does work against the time shifted replay, yep! A convincing `0.9976412969068766`.

kszlim: Nice, I guess the zero padding is robust after the inputs are normalized.

tybug:
this won't see use in circleguard (maybe ever) but it's nice to have it! Thanks for your initiative on this. I'll deal with the numpy/scipy issue another time (likely whenever something breaks on that front).

tybug:
I take that back - this will probably be the direction to go when offering a more thorough testing through circleguard, as it works well on t and xy shifts (and combinations thereof), which is more powerful than anything we have currently.

kszlim:
Nice, was a fun one.

PR #140, opened 2020-05-03
clean up tests

tybug:
* adds a (gitignored) cache to the tests
* splits tests into four files, three test files and a utils file

Should help highlight what we aren't testing currently (eg relax in `test_investigations`)

prompted largely by #139, will merge into that branch after this gets merged into master (unless you object @kszlim)


kszlim:
Sounds good to me!

PR #141, opened 2020-05-05
add a few type hints

tybug:
my linter (pylint) / IDE (vscode) don't seem to respect these type hints, so they're not useful unless someone is looking at the source, in which case we already document the return type. 

I'm curious if other editors treat type hints with more respect, though. Eg erroring when you do something contradictory to the type hint.

### DIFF
--- a/circleguard/circleguard.py
+++ b/circleguard/circleguard.py
@@ -5,7 +5,7 @@ import os
 from os.path import isfile, join
 import logging
 from tempfile import TemporaryDirectory
-from typing import Generator
+from typing import Iterable
 
 from circleguard.loader import Loader
 from circleguard.comparer import Comparer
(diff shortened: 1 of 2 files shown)


circleguard/circleguard.py
```diff
@@ -129,7 +132,7 @@ def run(self, loadables, detect, loadables2=None, max_angle=DEFAULT_ANGLE, min_d
                     # disconnect from temporary library
                     library.close()
 
-    def steal_check(self, loadables, loadables2=None):
+    def steal_check(self, loadables, loadables2=None) -> Generator[StealResult, None, None]:
```
InvisibleSymbol, line 135:
```suggestion
    def steal_check(self, loadables, loadables2=None) -> Iterable[StealResult]:
```
Should do the same thing but looks cleaner.

tybug: I was surprised to see that the [python docs](https://docs.python.org/3/library/typing.html#typing.Generator) recommend typing as an `Iterable` if your generator only yields

tybug:
Very happy to see that. Means that annotating our public api isn't a total waste.

PR #142, opened 2020-05-06
don't decompress cached replays twice

tybug:
roughly a 100% speedup (is that the right way to say it's twice as fast now? I think so). Nothing crazy but a very welcome change.


tybug:
ignore the failing tests, will be fixed when #142 is merged + released (this pr is blocked by that anyway)

PR #149, opened 2020-06-06
Average frametime

sometimeszzzzzzz:
a much more accurate average frametime calc

### DIFF
--- a/circleguard/investigator.py
+++ b/circleguard/investigator.py
@@ -46,7 +46,7 @@ class Investigator:
             snaps = self.aim_correction(self.replay, self.max_angle, self.min_distance)
             yield CorrectionResult(self.replay, snaps)
         if self.detect & Detect.TIMEWARP:
-            frametime = self.median_frametime(self.replay)
+            frametime = self.average_frametime(self.replay)
             yield TimewarpResult(self.replay, frametime)
 
     @staticmethod
@@ -195,7 +195,7 @@ class Investigator:
         return [jerks, ischeat]
 
     @staticmethod
-    def median_frametime(replay):
+    def _filter_frametime(replay):
         """
         Calculates the median time between the frames of ``replay``.
 
@@ -210,7 +210,45 @@ class Investigator:
         """
         # replay.t is cumsum so convert it back to "time since previous frame"
         t = np.diff(replay.t)
-        return np.median(t)
+        # placeholder 0, so t matches k
+        t = np.concatenate(([0], t))
+
+        filtered_frametimes = []
+        previous_key = 0
+        combine_next = False
+        for i in range(len(t)):
+            k = replay.k[i]
+            if previous_key == replay.k[i]:
+                if combine_next == False:
+                    filtered_frametimes.append(t[i])
+                else:
+                    filtered_frametimes.append(t[i-1] + t[i])
+                combine_next = False
+            else:
+                combine_next = True
+            previous_key = k
+        filtered_frametimes = filtered_frametimes[1:]
+        return np.median(filtered_frametimes), filtered_frametimes
+
+    @staticmethod
+    def average_frametime(replay):
+        """
+        Attempt at a more precise average frametime.
+        """
+        median, frametimes = Investigator._filter_frametime(replay)
+        filtered_frametimes = []
+        combine_next = False
+        for i in range(len(frametimes)):
+            if combine_next:
+                combine_next = False
+                filtered_frametimes.append(frametimes[i-1] + frametimes[i])
+            else:
+                if np.abs(median - frametimes[i]) > 3:
+                    combine_next = True
+                else:
+                    combine_next = False
+                    filtered_frametimes.append(frametimes[i])
+        return np.average(filtered_frametimes)
 
     @staticmethod
     def _parse_beatmap(beatmap):
--- a/circleguard/investigator.py
+++ b/circleguard/investigator.py
@@ -241,7 +241,8 @@ class Investigator:
         for i in range(len(frametimes)):
             if combine_next:
                 combine_next = False
-                filtered_frametimes.append(frametimes[i-1] + frametimes[i])
+                if np.abs(median-(frametimes[i-1] + frametimes[i])) <= 3:
+                    filtered_frametimes.append(frametimes[i-1] + frametimes[i])
             else:
                 if np.abs(median - frametimes[i]) > 3:
                     combine_next = True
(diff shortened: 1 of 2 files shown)


circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+        unique, counts = np.unique(t, return_counts=True)
+        unique = unique[counts > 4]
+        counts = counts[counts > 4]
+
+        # placeholder 0, so t matches k
+        t = np.concatenate(([0], t))
+        replay.k = replay.k[np.isin(t, unique)]
+        t = t[np.isin(t, unique)]
+
+        def mode(x):
```
tybug, line 225: use scipy.stats.mode

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+            # filters out low frametimes and grabs frametimes around the peaks
+            counts_cutoff = max(counts)/4
+            frametime_cutoff = 6
+            number_a = 7
+            if replay.mods.value & 1 << 6:
+                frametime_cutoff *= 1.5
+                number_a *= 1.5
+            if replay.mods.value & 1 << 8:
+                frametime_cutoff *= 0.75
+                number_a *= 0.75
```
tybug, line 340: should be using `replay.mods & Mod.DT` instead of this low level manipulation, but also this should use `utils.convert_statistic` which I believe is exactly this calculation

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+
+        # filters out insignificant frametimes
+        unique, counts = np.unique(t, return_counts=True)
+        unique = unique[counts > 4]
+        counts = counts[counts > 4]
+
+        # placeholder 0, so t matches k
+        t = np.concatenate(([0], t))
+        replay.k = replay.k[np.isin(t, unique)]
+        t = t[np.isin(t, unique)]
```
tybug, line 223: this is filtering out frames where the frametime appears less than 4 times in the replay? Would prefer to do this by removing outliers instead, using a mathematical definition (eg more than 2 stddevs away from the mean)

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+            m = counts.argmax()
+            return values[m]
+
+        filtered_frametimes = []
+        previous_key = 0
+        combine_next = False
+        # for very fast buzz sliders
+        count_lowframes = 0
+        current_mode = 5
+        for i in range(len(t)):
```
tybug, line 238: use `for i, curr_t in enumerate(t)` instead

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+                    combined_time = t[i-1] + t[i]
+                    if combined_time > current_mode + 5:
+                        filtered_frametimes.extend([t[i-1], t[i]])
+                    else:
+                        filtered_frametimes.append(combined_time)
+                combine_next = False
+            else:
+                combine_next = True
+            previous_key = k
+        filtered_frametimes = filtered_frametimes[1:]
```
tybug, line 267: I'm confused as to what this entire thing is doing. Why does mode start at 5? What are these magic numbers doing here (`combined_time > current_mode + 5`, `count_lowframes > 8`)? Please extract them out as constants.

There's also many indentation levels and little documentation. When something is this complicated, please extract it to a new method(s) with a docstring.

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+        
+        # filters 3 times to filter out more small frametimes
+        for i in range(3):
```
tybug, line 270: is this a different filtering than what you just did above? If so, why apply two different filtering algorithms?

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+            for i in range(len(filtered_frametimes)):
+                if combine_next:
+                    combine_next = False
+                    combined_time = filtered_frametimes[i-1] + filtered_frametimes[i]
+                    if combined_time > median + 5:
+                        filtered_frametimes_more.extend([filtered_frametimes[i-1], filtered_frametimes[i]])
+                    else:
+                        filtered_frametimes_more.append(combined_time)
+                else:
+                    if np.abs(median - filtered_frametimes[i]) > 4:
```
tybug, line 283: magic number..

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+
+        median, frametimes, lowhz_keyboard = Investigator._filter_frametime(replay)
+        unique, counts = np.unique(frametimes, return_counts=True)
+
+        # exception for 100hz polling keyboards
+        if lowhz_keyboard:
+            # removes some lower frametimes that were likely missed during filtering
+            counts_cutoff = max(counts)/6
+            remove_index = []
+            for i in range(len(counts)):
```
tybug, line 322: use enumerate here

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+        avg frametime < 13 is 100% timewarp
+        """
+
+        median, frametimes, lowhz_keyboard = Investigator._filter_frametime(replay)
+        unique, counts = np.unique(frametimes, return_counts=True)
+
+        # exception for 100hz polling keyboards
+        if lowhz_keyboard:
+            # removes some lower frametimes that were likely missed during filtering
+            counts_cutoff = max(counts)/6
```
tybug, line 320: magic number

circleguard/investigator.py
```diff
@@ -206,11 +206,153 @@ def median_frametime(replay):
...
+            counts = np.delete(counts, remove_index)
+            unique = np.delete(unique, remove_index)
+            filtered_frametimes = frametimes[np.isin(frametimes, unique)]
+            # increase by a bit since replays with low fps tend to be a bit lower (due to shorter frametimes being a bit harder to accurately filter out)
+            return np.average(filtered_frametimes) + 0.1
+        else:
+            # filters out low frametimes and grabs frametimes around the peaks
+            counts_cutoff = max(counts)/4
+            frametime_cutoff = 6
+            number_a = 7
```
tybug, line 334: what is `number_a` and why is it 7? The name tells me nothing and the value doesn't help...

PR #153, opened 2020-06-25
fix frametimes for relax plays

sometimeszzzzzzz:
removes consecutive short frametimes caused by relax mod and switching desktops during a break

### DIFF
--- a/circleguard/investigator.py
+++ b/circleguard/investigator.py
@@ -207,7 +207,12 @@ class Investigator:
             The replay to get the frametimes of.
         """
         # replay.t is cumsum so convert it back to "time since previous frame"
-        return np.diff(replay.t)
+        t = np.diff(replay.t)
+        # remove low frametimes caused by relax mod and switching desktops during a break
+        t = Investigator._remove_low_frametimes(t, 4)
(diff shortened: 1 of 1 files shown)


circleguard/investigator.py
```diff
@@ -264,6 +268,20 @@ def _filter_hits(hitobjs, keypress_times, OD):
 
         return array
 
+    @staticmethod
+    def _remove_low_frametimes(frametimes, limit):
+        cnt = 0
+        lowframes = []
+        for i, curr_t in enumerate(frametimes):
```
tybug, line 275: what's with the brevity? No need to abbreviate things like this. Use `frametime` instead of `curr_t`, `num_low_frametimes` instead of `cnt`, and `low_frametimes` instead of `lowframes`

tybug:
I think the method could be written a little more elegantly but I like the direction. Can you upload a few relax plays to `test/resources/legit` (and make that folder)? Ideally we'd have a play which is +RX *and* timewarped, but that seems hard to come by.

I also made this only run on relax plays. We might want to filter all replays in a similar fashion in the future, but I'd rather see how common that is without relax first.

I'm also unsure about filtering *all* 0 frame times. I would think that just looking at runs of short time frames would be enough.

PR #157, opened 2020-07-23
Improve hit error filtering/ur calc

sometimeszzzzzzz:
- [ ]  depends on https://github.com/llllllllll/slider/pull/67

### DIFF
--- a/circleguard/investigator.py
+++ b/circleguard/investigator.py
@@ -76,7 +76,6 @@ class Investigator:
 
         version = replay.game_version
 
-
         if version is None:
             # estimate version with timestamp, this is only accurate if the user
             #  keeps their game up to date. We don't get the timestamp from
--- a/circleguard/investigator.py
+++ b/circleguard/investigator.py
@@ -4,6 +4,8 @@ import numpy as np
 
 from circleguard.enums import Key, Detect
 from circleguard.result import RelaxResult, CorrectionResult, TimewarpResult
+from circleguard.mod import Mod
+from slider.beatmap import Circle, Slider, Spinner
 
 class Investigator:
     """
@@ -26,6 +28,7 @@ class Investigator:
     """
 
     MASK = int(Key.M1) | int(Key.M2)
+    VERSION_SLIDERBUG_FIXED = 20190513
 
     def __init__(self, replay, detect, max_angle, min_distance, beatmap=None):
         self.replay = replay
@@ -64,12 +67,13 @@ class Investigator:
             The beatmap to calculate ``replay``'s ur with.
         """
 
-        easy = replay.mods.value & 1 << 1
-        hard_rock = replay.mods.value & 1 << 4
+        easy = Mod.EZ in replay.mods
+        hard_rock = Mod.HR in replay.mods
         hitobjs = Investigator._parse_beatmap(beatmap, easy, hard_rock)
         version = replay.game_version
 
-        # use the timestamp as the version if it doesn't exist
+        # replicate version with timestamp, this is only accurate if the user keeps their game up to date.
+        # We don't get the timestamp from `ReplayMap`, only `ReplayPath`
         if version == None:
             version = int(f"{replay.timestamp.year}{replay.timestamp.month:02d}{replay.timestamp.day:02d}")
 
@@ -241,19 +245,7 @@ class Investigator:
 
         # parse hitobj
         for hit in beatmap.hit_objects(easy=easy, hard_rock=hard_rock):
-            time = hit.time.total_seconds() * 1000
-            end_time = time
-            obj_type = 0
-            # object end time is different depending on the type of object
-            # type of hit object is necessary for notelock calculations (0 circle, 1 slider, 2 spinner)
-            if hasattr(hit, "end_time"):
-                end_time = hit.end_time.total_seconds() * 1000
-                if hasattr(hit, "curve"):
-                    obj_type = 1
-                else:
-                    obj_type = 2
-            p = hit.position
-            hitobjs.append([time, [p.x, p.y], end_time, obj_type])
+            hitobjs.append(hit)
         return hitobjs
 
     @staticmethod
@@ -299,17 +291,26 @@ class Investigator:
 
 
         while object_i < len(hitobjs) and press_i < len(replay_data):
-            hitobj_time = hitobjs[object_i][0]
+            hitobj = hitobjs[object_i]
+
+            hitobj_time = hitobj.time.total_seconds() * 1000
             press_time = replay_data[press_i][0]
 
-            hitobj_xy = hitobjs[object_i][1]
+            hitobj_xy = [hitobj.position.x, hitobj.position.y]
             press_xy = replay_data[press_i][1]
 
-            hitobj_end_time = hitobjs[object_i][2]
-            hitobj_type = hitobjs[object_i][3]
+            if isinstance(hitobj, Circle):
+                hitobj_type = 0
+                hitobj_end_time = hitobj_time
+            elif isinstance(hitobj, Slider):
+                hitobj_type = 1
+                hitobj_end_time = hitobj.end_time.total_seconds() * 1000
+            else:
+                hitobj_type = 2
+                hitobj_end_time = hitobj.end_time.total_seconds() * 1000
 
             # before sliderbug fix, notelock ended after hitwindow50
-            if version < 20190513:
+            if version < Investigator.VERSION_SLIDERBUG_FIXED:
                 notelock_end_time = hitobj_time + hitwindow
                 # exception for sliders/spinners, where notelock ends after hitobject end time if it's earlier
                 if hitobj_type != 0:
@@ -337,7 +338,7 @@ class Investigator:
                     # sliders don't disappear after missing
                     # so we skip to the press_i that is after notelock_end_time
                     press_i += 1
-                    if press_i < len(replay_data) and hitobj_type == 1 and version >= 20190513:
+                    if press_i < len(replay_data) and hitobj_type == 1 and version >= Investigator.VERSION_SLIDERBUG_FIXED:
                         while replay_data[press_i][0] <= notelock_end_time:
                             press_i += 1
                             if press_i >= len(replay_data):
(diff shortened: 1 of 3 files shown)


circleguard/investigator.py
```diff
@@ -64,9 +64,19 @@ def ur(replay, beatmap):
             The beatmap to calculate ``replay``'s ur with.
         """
 
-        hitobjs = Investigator._parse_beatmap(beatmap)
-        keypress_times = Investigator._parse_keypress_times(replay)
-        filtered_array = Investigator._filter_hits(hitobjs, keypress_times, beatmap.overall_difficulty)
+        easy = replay.mods.value & 1 << 1
```
tybug, line 67: use `Mod.EZ in replay.mods`, same for hr

circleguard/investigator.py
```diff
@@ -226,42 +236,137 @@ def median_frametime(frametimes):
...
 
         # parse hitobj
-        for hit in beatmap.hit_objects_no_spinners:
+        for hit in beatmap.hit_objects(easy=easy, hard_rock=hard_rock):
+            time = hit.time.total_seconds() * 1000
+            end_time = time
+            obj_type = 0
+            # object end time is different depending on the type of object
+            # type of hit object is necessary for notelock calculations (0 circle, 1 slider, 2 spinner)
+            if hasattr(hit, "end_time"):
```
tybug, line 249: don't use attr checks ever if you can avoid it. Check against the instance of `hit` - `if isinstance(hit, Slider)`, etc

circleguard/investigator.py
```diff
@@ -226,42 +236,137 @@ def median_frametime(frametimes):
...
         array = []
-        hitwindow = 150 + 50 * (5 - OD) / 5
+
+        # stable converts the OD, which is originally a float32, to a double
+        # and this causes some hitwindows to be messed up when casted to an int
+        # so we replicate this
+        hitwindow = int(150 + 50 * (5 - float(np.float32(OD))) / 5) - 0.5
+
+        # attempting to match stable hitradius
+        hitradius = np.float32(64 * ((1.0 - np.float32(0.7) * (float(np.float32(CS)) - 5) / 5)) / 2) * np.float32(1.00041)
```
tybug, line 295: formula source?

circleguard/investigator.py
```diff
@@ -226,42 +236,137 @@ def median_frametime(frametimes):
...
-            elif press_time > hitobj_time + hitwindow / 2:
-                object_i += 1
+            hitobj_xy = hitobjs[object_i][1]
+            press_xy = replay_data[press_i][1]
+
+            hitobj_end_time = hitobjs[object_i][2]
+            hitobj_type = hitobjs[object_i][3]
+
+            # before sliderbug fix, notelock ended after hitwindow50
+            if version < 20190513:
```
tybug, line 312: extract out to static class variable (`VERSION_SLIDERBUG_FIXED`)

circleguard/investigator.py
```diff
@@ -64,9 +64,19 @@ def ur(replay, beatmap):
...
 
-        hitobjs = Investigator._parse_beatmap(beatmap)
-        keypress_times = Investigator._parse_keypress_times(replay)
-        filtered_array = Investigator._filter_hits(hitobjs, keypress_times, beatmap.overall_difficulty)
+        easy = replay.mods.value & 1 << 1
+        hard_rock = replay.mods.value & 1 << 4
+        hitobjs = Investigator._parse_beatmap(beatmap, easy, hard_rock)
+        version = replay.game_version
+
+        # use the timestamp as the version if it doesn't exist
```
tybug, line 72:
```suggestion
        # replicate version with timestamp, this is only accurate if the user keeps their game up to date.
        # We don't get the timestamp from `ReplayMap`, only `ReplayPath`
```

circleguard/investigator.py
```diff
@@ -226,42 +236,137 @@ def median_frametime(frametimes):
...
+            # type of hit object is necessary for notelock calculations (0 circle, 1 slider, 2 spinner)
+            if hasattr(hit, "end_time"):
+                end_time = hit.end_time.total_seconds() * 1000
+                if hasattr(hit, "curve"):
+                    obj_type = 1
+                else:
+                    obj_type = 2
             p = hit.position
-            hitobjs.append([hit.time.total_seconds() * 1000, p.x, p.y])
+            hitobjs.append([time, [p.x, p.y], end_time, obj_type])
```
tybug, line 256: This should really become its own class, too much magic indexing going on where I don't know what `hitobjs[i][0]` means. `hitobjs[i].time` is clearer

sometimeszzzzzzz: I think I'll turn it into a list of hitobjects instead, since this part is basically removing attributes of hitobject that I'm not using

tybug: yep that's better

circleguard/investigator.py
```diff
@@ -226,42 +240,134 @@ def median_frametime(frametimes):
...
+                    filtered_replay_data.append([replay.t[i], replay.xy[i], False])
+
+        # add a duplicate frame when 2 keys are pressed at the same time
+        changes = changes[changes != 0]
+        i = 0
+        for j in np.where(changes == 3)[0]:
+            filtered_replay_data.insert(j + i + 1, filtered_replay_data[j + i])
+            i += 1
+        
+        return np.array(filtered_replay_data, dtype=object)
```
tybug, line 274: is it necessary to convert this to a numpy array?

sometimeszzzzzzz: probably not

circleguard/investigator.py
```diff
@@ -226,41 +242,146 @@ def median_frametime(frametimes):
...
+                        while replay_data[press_i][0] <= notelock_end_time:
+                            press_i += 1
+                            if press_i >= len(replay_data):
+                                break
+                    object_i += 1
+                # keypress not on object, so we move to the next keypress
+                else:
+                    press_i += 1
 
         return array
```
tybug, line 387: I expect hits to be useful for things other than UR, so we should make a new class `Hit` for it which contents the replay data at that frame (or equivalent information in the form of t, xy, and k), the hitobject that got hit, and maybe anything else that's relevant

This is more of a note for myself but if you get to it first feel free

circleguard/investigator.py
```diff
@@ -26,6 +28,8 @@ class Investigator:
     """
 
     MASK = int(Key.M1) | int(Key.M2)
+    # https://osu.ppy.sh/home/changelog/cuttingedge/20190111
+    VERSION_SLIDERBUG_FIXED = 20190111
```
sometimeszzzzzzz, line 32: I asked MillhioreF and if MillhioreF is correct, the initial implementation of sliderbug fix was this version. Will have to do some testing later to confirm.

tybug: only for cutting edge though? we should probably be using the first version it was fixed for stable (maybe with a comment to the first cutting edge changelog where it was fixed)

sometimeszzzzzzz: stable implementation of this was 20190207, but the previous stable version before that is 20181231, so using cutting edge version number won't affect plays set on stable version 20181231.

tybug: I see, makes sense

tybug: hmm tough call with the versioning...I think people use circleguard more often with non-local replays, so we should probably err on the side of caution (ie 20190207) so we don't give the wrong ur for stable replays set between 20190111 and 20190207

sometimeszzzzzzz: yeah I was thinking that too, <3% of players use cutting edge anyways. I could add an exception for local replays to use the cutting edge version, not sure if that's a good idea though.

tybug: sure, that's probably fine. `Replay` subclasses are stable enough I'm not too concerned about relying on their implementation details (like having a concrete version or not)

circleguard/investigator.py
```diff
@@ -226,43 +250,154 @@ def median_frametime(frametimes):
...
+                # so we need to keep track of the frames where this happens
+                if relative_t[i] == 0:
+                    keydown_frames.append([replay.t[i], replay.xy[i], True])
+                else:
+                    keydown_frames.append([replay.t[i], replay.xy[i], False])
+
+        # add a duplicate frame when 2 keys are pressed at the same time
+        keydowns = keydowns[keydowns != 0]
+        i = 0
+        for j in np.where(keydowns == 3)[0]:
```
tybug, line 294: why exactly 3? I assume this is `Key.M1 + Key.M2`, but this will ignore eg when both `Key.K1` and `Key.K2` are pressed. Shouldn't it be `keydowns & (Key.M1 + Key.M2) == (Key.M1 + Key.M2)`? (works for when all four bits are on, aka K1 and K2 are pressed, which are always set with M1 and M2 respectively - 15 & 3 == 3, and works for M1 and M2 as well, 3 & 3 == 3)

circleguard/investigator.py
```diff
@@ -226,43 +250,154 @@ def median_frametime(frametimes):
...
+                hitobj_i += 1
+            else:
+                if keydown_t < hitobj_t + hitwindow and np.linalg.norm(keydown_xy - hitobj_xy) <= hitradius and hitobj_type != 2:
+                    hits.append([hitobj_t, keydown_t])
+
+                    # sliders don't disappear after clicking
+                    # so we skip to the press_i that is after notelock_end_time
+                    keydown_i += 1
+                    if keydown_i < len(keydowns) and hitobj_type == 1 and version >= version_sliderbug_fixed:
+                        while keydowns[keydown_i][0] <= notelock_end_time:
```
tybug, line 391: you use `>= notelock_end_time` and ` <= notelock_end_time`, which is it? I don't think they can both be true; either a keydown at exactly the time notelock ends interacts with the note, or it doesn't

circleguard/investigator.py
sometimeszzzzzzz: I think `keypresses = replay.k & Investigator.MASK` already converts Key.K1 and Key.K2 to Key.M1 and Key.M2 respectively

tybug: oh, well spotted. You're right. Let's still compare keydowns to MASK then - `keydowns == Investigator.MASK`, to remove the magic number 3

circleguard/investigator.py
sometimeszzzzzzz: thx for noticing that, completely missed it. I'm actually not sure which one it is. Will do some tests later, since I also need to figure out if the game rounds slider end time to fix this issue.

PR #159, opened 2020-08-07
add ReplayString class

### LINKED ISSUE #158: Loading and processing a replay from memory
It would be helpful to add a way to load the replay directly from memory. For example, load a file from s3 and transfer it to circlecore without saving a temporary file.

tybug:
usage:

```python
replay_data_str = open("replay.osr", "rb").read()
r = ReplayString(replay_data_str)
cg.load(r)
print(cg.ur(r, single=True).ur)
```

closes #158 

@deissh please confirm this addresses your use case (if you're not sure just by looking at it, check out this branch locally and try for yourself)


deissh:
awesome 👍

tybug:
should be available in core 4.4.1

PR #163, opened 2020-12-14
add lots of replays which produce false positives or false negatives

myangelaku:
I'm still learning how to use GitHub properly so please don't blame me if this is messy >.<


tybug:
after the recent addition of filtering snaps to only contain those on hitobjs:
* 35723e4a932fb37843fcbe5270e0e16ebec15eee has 0 snaps 
* c0682ecaa7639aabcec44cb4c44bae8596f29021 has 0 snaps
* 4332bda14ece17f6daa4f3c824b9bb93086cf354 has 1 snap

rest of them need more investigation on my part so I can add tests for them / see where to improve (don't remove the above replays  though, will add tests for them to ensure we don't regress in behavior).

tybug:
Thanks for your work on this, very helpful collection of testing replays.

PR #166, opened 2021-03-04
Fixes lowercase mods not being registered

hxotic:
6zmfR789
^^ fixes this im pretty sure


tybug:
I think I want to keep requiring the string to be uppercase in circlecore actually. This should be changed in circleguard, so uppercasing the string before passing it to `Mod(mod_string)`. It should be changed on any line that looks like this: https://github.com/circleguard/circleguard/blob/master/circleguard/widgets.py#L582, of which there are 6 lines total, across 3 loadables (I can do this if you don't want to).

PR #171, opened 2021-04-15
Add false positives to tests

myangelaku:


tybug:
ignore the failing tests, will be fixed by a new slider release

(this is why you don't merge stuff to master before its upstream dependencies have been merged/released, but who's going to stop me)

myangelaku:
Please note that the reason why I put 2 pairs of replays for each map is because the lowest results seems to be from replays from the same player; I don't know if that changes anything so I'll also upload the lowest similarity results of different players if that's the case

myangelaku:
I think one per map and player is enough

myangelaku:
Context for Windows Ink Snaps (well known due to [the case of Mahluna](https://www.reddit.com/r/osureport/comments/m557lr/osustd_mahluna_blatant_aim_correction/)):

https://youtu.be/MHFfZh_4nK4
https://youtu.be/T4I2taoVem8

myangelaku:
For explanation: 
I'll consider every replay that has lower UR than [the current record which is at 35,75](https://www.reddit.com/r/osugame/comments/8lqcyh/new_osustandard_ur_record_by_corim/) blatant.

tybug:
folders should have underscores instead of spaces

myangelaku:
> 
> 
> folders should have underscores instead of spaces

Right, will fix!

PR #176, opened 2021-09-15
Nightcore?

Skylocc:


circleguard/frametime_graph.py
```diff
@@ -94,6 +94,8 @@ def plot_with_break(self, frametimes):
     def _conversion_factor(self, replay):
         if Mod.DT in replay.mods:
```
Stedoss, line 95: I guess it would be fine (and a little more performant) if you just add it to the first conditional.
```suggestion
        if Mod.DT or Mod.NC in replay.mods:
```

tybug:
Thanks for the contribution, but DT is always set with NC, so this is already accounted for:

```python
from circleguard import Mod
assert Mod.DT in Mod.NC
```

PR #180, opened 2022-04-18
Added functionality to calculate corrected UR

thedarklort:
Added a function to calculate an advanced version of UR in order to determine more correctly on how spread Hits are on replays

### DIFF
--- a/circleguard/investigations.py
+++ b/circleguard/investigations.py
@@ -19,7 +19,7 @@ class Investigations:
     VERSION_SLIDERBUG_FIXED_CUTTING_EDGE = GameVersion(20190111, concrete=True)
 
     @staticmethod
-    def ur(replay, beatmap):
+    def ur(replay, beatmap, adjusted=False):
         """
         Calculates the ur of ``replay`` when played against ``beatmap``.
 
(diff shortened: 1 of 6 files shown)


circleguard/investigations.py
```diff
@@ -36,6 +36,23 @@ def ur(replay, beatmap):
         diffs = [hit.error() for hit in hits]
         return np.std(diffs) * 10
 
+    @staticmethod
+    def correctedur(replay, beatmap):
```
tybug, line 40: I'd prefer this to be implemented as an argument `corrected` in the `ur` function, defaulting to `False`, instead of an entirely new function

circleguard/utils.py
```diff
@@ -220,6 +220,42 @@ def hitradius(CS):
     return np.float32(64 * ((1.0 - np.float32(0.7) * (float(np.float32(CS)) - 5) / 5)) / 2) * np.float32(1.00041)
 
 
+def GetQuartiles(arr):
```
tybug, line 223: use snake case for function and variable names please https://peps.python.org/pep-0008/#function-and-variable-names

circleguard/utils.py
```diff
@@ -220,6 +220,42 @@ def hitradius(CS):
...
+        List of hits
+    bias = numeric:`number`
+        How high the bias should be (standard is 1.5)
+
+    """
+    q1,q3 = GetQuartiles(arr)
+    iqr = q3 - q1
+    lowerLimit = q1 - bias*iqr
+    upperLimit = q3 + bias*iqr
+    return filter(lambda x: (x > lowerLimit and x < upperLimit),arr)
```
tybug, line 256: I prefer writing filters with list comprehensions, so this becomes

`return [x for x in arr if x lowerLimit < x < upperLimit]`

(but also as above please use snake case for variables)

tybug (changes requested):

circleguard/investigations.py
```diff
@@ -19,7 +19,7 @@ class Investigations:
     VERSION_SLIDERBUG_FIXED_CUTTING_EDGE = GameVersion(20190111, concrete=True)
 
     @staticmethod
-    def ur(replay, beatmap):
+    def ur(replay, beatmap, adjusted=False):
```
tybug, line 22: it's good that you changed this function, but your work isn't done yet: consumers don't actually call `Investigator.ur`, so this change is useless right now. You need to modify `Circleguard.ur` to take an `adjusted` parameter as well.

circleguard/utils.py
```diff
@@ -220,6 +220,43 @@ def hitradius(CS):
...
+
+    Parameters
+    ----------
+    arr: list:`numeric list`
+        List of hits
+    bias = numeric:`number`
+        How high the bias should be (standard is 1.5)
+
+    """
+    q1,q3 = get_quartiles(arr)
```
tybug, line 253: instead of writing our own `get_quartiles` function, I would use `np.percentile(x, [75 ,25])` to get the 75th and 25th percentile (q3 and q1 respectively I believe?): https://stackoverflow.com/a/23229224/12164878

tybug:
I've tidied up the docs and added tests. This looks good. The difference between adjusted ur and normal ur is slightly worrying at first glance (as much as a 35% decrease):

image

but I think it's fine since the replays in question do have some pretty bad outliers which I would expect to throw off the ur a fair bit.

PR #181, opened 2024-02-16
fix: return entire array if no outliers are found

gengotech:
fix in the `filter_outliers` method, with previous behavior if no outliers were found, an empty array would have been returned. this results in adjusted ur calculation being `None` in some replays (while normal ur calculation would be fine)

this conflicts with the usage and documentation of the method, which states that it'd return the array *with* the outliers removed.

to fix this, we simply return the original array if no outliers were found.

### DIFF
--- a/circleguard/utils.py
+++ b/circleguard/utils.py
@@ -235,8 +235,8 @@ def filter_outliers(arr, bias=1.5):
     iqr = q3 - q1
     lower_limit = q1 - (bias * iqr)
     upper_limit = q3 + (bias * iqr)
-    outliers = [x for x in arr if lower_limit < x < upper_limit]
-    return arr if not outliers else outliers
+    arr_without_outliers = [x for x in arr if lower_limit < x < upper_limit]
+    return arr if not arr_without_outliers else arr_without_outliers
 
 
 TRACE = 5


tybug:
am I misreading or doesn't this return the entire array if *everything* is an outlier? if nothing if an outlier, then this function still returns the entire array.

I think it's actually somewhat valid to return an invalid adjusted ur when everything is an outlier (because it's very unreliable if everything is an outlier), but open to opinions.

tybug:
put another way: this function in master already does what this pr says, which is return the entire array when there are no outliers. so something is being lost in translation here

tybug:
> skipping semantics, as it is in master an .ur(adjusted=True) calculation will return None if there are no outliers

do you have an example? I can't reproduce this.

tybug:
That's a case where *everything* is an outlier, not nothing.

fwiw, I get `nan` and an error for normal and adjusted ur respectively for that replay:

```python
r = cg.ReplayMap(1711983, 33839341)
print(cg.ur(r)) # nan
print(cg.ur(r, adjusted=True)) # error
```

```
...
  File "/opt/homebrew/lib/python3.12/site-packages/numpy/lib/function_base.py", line 4830, in _quantile
    slices_having_nans = np.isnan(arr[-1, ...])
                                  ~~~^^^^^^^^^
IndexError: index -1 is out of bounds for axis 0 with size 0
```

I'm fine with including the endpoints for outlier filtering to avoid this in the case of 0 iqr. We also probably want to raise an error when calculating the ur of a replay with < 3 hits (I thought we already did, to be honest).

PR #182, opened 2024-02-20
Replace Appveyor PR pipeline with Github Actions

Stedoss:
Super simple pipeline to hopefully replace the older one.

Python 3.9 was picked because it worked, but we can pick another or do a matrix (eg. `3.9`, `3.10`, `3.11`).

Have also updated the tests to reflect the current state of the leaderboards for the tests, but they can be changed/removed if we want.

Requirements for merge:
* API Key needs to be added to the repo actions secrets.


tybug:
just read this, which might mean this idea is dead in the water....emphasis mine https://github.com/circleguard/circlecore/settings/secrets/actions

> Anyone with collaborator access to this repository can use these secrets and variables for actions. **They are not passed to workflows that are triggered by a pull request from a fork.**

this is a huge test improvement regardless though, thanks! appveyor definitely needed to go.

